PR #51, opened 2022-08-01
Feature/improve integration test

aymeric-dispa:
- Add integration tests in a new module 'integrationTest'
- Fix small bugs (property handling)
- Improve exception handling (404)
- Rename/refactor some classes to improve code cleanliness and reusibility (eg:  renamed FireboltDatabaseMetadataResult to QueryResult so that its usage is not only limited to Metadata)
- Add common test modules 'testCommon'


src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
```diff
@@ -32,7 +32,7 @@ public enum JavaTypeToFireboltSQLString {
 	ARRAY(Array.class, SqlArrayUtil::arrayToString);
 
 	private static final Pair<String[], String[]> characterToEscapedCharacterPair = new ImmutablePair<>(
-			new String[] { "\0", "\\", "'", "`" }, new String[] { "\\0", "\\\\", "\\'", "\\`" });
+			new String[] { "\0", "\\", "'" }, new String[] { "\\0", "\\\\", "\\'" });
```
ptiurin, line 35: Why's this no longer escaped?

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -134,6 +133,8 @@ private void abortStatementRunningOnFirebolt(String statementId) throws SQLExcep
 				statementService.abortStatement(statementId, this.sessionProperties);
 			}
 			log.debug("Statement with id {} was aborted", statementId);
+		} catch (FireboltException e) {
+			throw e;
```
ptiurin, line 137: is this not redundant?

src/integrationTest/java/integration/tests/StatementCancelTest.java
```diff
@@ -0,0 +1,106 @@
...
+			rs.close();
+		}
+
+	}
+
+	private String extractTableName() throws SQLException {
+		String tableName = null;
+		try (Connection connection = this.createConnection(); Statement statement = connection.createStatement()) {
+			this.setParam(connection, "use_standard_sql", "0");
+			ResultSet resultSet = statement.executeQuery("SHOW tables");
```
ptiurin, line 93: I think `SHOW tables` works without use_standard_sql override.

ptiurin: Same in some other places.

ptiurin (commented):
Generally lgtm, just some questions.

src/integrationTest/java/integration/tests/DatabaseMetaDataTest.java
```diff
@@ -0,0 +1,129 @@
...
+	@Test
+	void shouldReturnSchema() throws SQLException {
+		List<String> schemas = new ArrayList<>();
+		List<String> catalogs = new ArrayList<>();
+		try (Connection connection = createConnection()) {
+			DatabaseMetaData databaseMetaData = connection.getMetaData();
+
+			try (ResultSet resultSet = databaseMetaData.getSchemas()) {
+				while (resultSet.next()) {
+					schemas.add(resultSet.getString("TABLE_SCHEM"));
```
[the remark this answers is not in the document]
aymeric-dispa: Noope, I know it's odd but getSchemas() is expected to returrn TABLE_SCHEM

src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
aymeric-dispa: I just give it a try and I couldnt use ` as a quote (it works as a standard character so I don't think we should escape it)

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
aymeric-dispa: No because just below we do 
catch (Exception e) {
....

We want to throw a FireboltException if an exception that is not a FireboltException is found. If the exception is already a FireboltException then we simply re-throw it

src/integrationTest/java/integration/tests/StatementCancelTest.java
ptiurin: Ok fair, let's pop a comment there so we remember that's how it works.

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
ptiurin: Ok I see

src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
ptiurin: Interesting, looks like in CH ` is equivalent to ", but we're not allowing it at all. Probably for the best.

ptiurin (approved):

src/integrationTest/java/integration/tests/StatementCancelTest.java
aymeric-dispa: ok 👍 I changed the class a bit so that it makes more sense (changed method name and I'm reusing the connection)

PR #54, opened 2022-08-04
Add CI config for release

aymeric-dispa:


.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: Release new version
+
+on:
+  release:
+    types: [published]
+
+env:
+  AWS_REGION: "us-east-1"
+  AWS_ACCOUNT_ID: "231290928314" # publishing
```
stepansergeevitch, line 9: Should this be stored in secrets? I can't tell if it's sensitive or not

stepansergeevitch (approved):
lgtm

.github/workflows/release.yml
aymeric-dispa: I will move it to secrets - better safe than sorry 👍

PR #61, opened 2022-08-15
Store tokens 5 seconds shorter than expiry time to avoid using token…

aymeric-dispa:
- Store tokens 30 seconds shorter than expiry time to avoid using tokens that are about to expire
- Improve unit tests

### DIFF
--- a/src/main/java/com/firebolt/jdbc/service/FireboltAuthenticationService.java
+++ b/src/main/java/com/firebolt/jdbc/service/FireboltAuthenticationService.java
@@ -26,6 +26,9 @@ public class FireboltAuthenticationService {
 			.variableExpiration().build();
 	private final FireboltAuthenticationClient fireboltAuthenticationClient;
 
+	private static final long TOKEN_EXPIRATION_OFFSET = 5L;
+	private static final long TOKEN_TTL_THRESHOLD = 60L;
+
 	public FireboltConnectionTokens getConnectionTokens(String host, FireboltProperties loginProperties)
 			throws FireboltException {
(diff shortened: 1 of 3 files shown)


src/main/java/com/firebolt/jdbc/service/FireboltAuthenticationService.java
```diff
@@ -49,6 +50,14 @@ public FireboltConnectionTokens getConnectionTokens(String host, FireboltPropert
 		}
 	}
 
+	/**
+	 * To avoid returning tokens that are about to expire, we store them 30 seconds shorter than their expiry time
+	 */
+	private long getCachingDurationInSeconds(long expireInSeconds) {
+		long thirtySeconds = 30L;
```
ptiurin, line 57: I'd probably make it a const outside of this function, for visibility.

ptiurin (commented):
How did you come up with 30s? What happens if a token expires while a long query is running?

aymeric-dispa:
> How did you come up with 30s? What happens if a token expires while a long query is running?

Usually tokens take hours before expiring. That change will simply make the tokens 'expire' 30 seconds early in the driver so that the client won't face an issue in case the token expires after it has been fetched from the local cache and before the query is sent sent our API. The driver does not cancel any query if a token gets expired (authorization is still performed by the backend - so there is no change regarding long queries). Tokens are fetched every time a new query is sent to our API (so in that regard 30 seconds is actually very safe)

ptiurin (approved):
apart from that lgtm

PR #76, opened 2022-09-06
Add support for service accounts

aymeric-dispa:
- Add support for service accounts
  - User can login the same way as with a username/password except that they have to previde the clientId/secret.
- Improve the error message when facing a 401 as it may happen if the user provides an invalid clientId/secret pair
- Change project version to 2.2.0 SNAPSHOT


stepansergeevitch (approved):
LGTM. The question is will token caching work fine with service accounts? Can we add a test?

aymeric-dispa:
> LGTM. The question is will token caching work fine with service accounts? Can we add a test?
Thanks 👍 
Yep it will work because the caching is done one layer above - so the method that calls the service to authenticate has no idea of wheter a username is an actual username or a clientId and will store the token regardless (it's only in the client methods that the logic is different). I could add a test but it is actually already tested since it's the same method (so the test would be testing the same thing) so I think it's better to keep it like that

PR #114, opened 2022-10-19
Build: Exclude slf4j from shading

spinscale:
If you take a look at the shaded jar, the log4j dependency is included, but not shaded.

Note: This makes #113 obsolete, as now the log4j dependency needs to be defined as a proper dependency, whereas that PR removes all dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -138,9 +138,11 @@ shadowJar {
     packages.each {
         relocate(it, "com.firebolt.shadow.${it}")
     }
-    // while slf4j is not relocated, but should still not be shaded
-    exclude("org/slf4j/**")
-    exclude("META-INF/maven/org.slf4j/**")
+    if (!project.hasProperty("addLogbackDependency")) {
+        // while slf4j is not relocated, but should still not be shaded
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -125,19 +125,22 @@ shadowJar {
...
             }
-            jf.close()
         }
     }
     packages.each {
         relocate(it, "com.firebolt.shadow.${it}")
     }
+    // while slf4j is not relocated, but should still not be shaded
+    exclude("org/slf4j/**")
+    exclude("META-INF/maven/org.slf4j/**")
```
aymeric-dispa, line 143:
```suggestion
    if (!project.hasProperty("addLogbackDependency")) {
        // while slf4j is not relocated, but should still not be shaded
        exclude("org/slf4j/**")
        exclude("META-INF/maven/org.slf4j/**")
    }
```
I think we still need to include it when the jar is used but SLF4J is not provided by the client (eg: for our integration tests)

spinscale:
Closing this for now, as this would mean to provide slf4j manually at runtime, which is not feasible...

PR #231, opened 2023-05-03
Update version to 2.4.4

yingli017:


build.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.firebolt'
-version = '2.4.3'
```
yingli017, line 12: not sure whether i should create `2.4.4-SNAPSHOT` or `2.4.4`?

ptiurin: Snapshot is generally pre-release, for changes that you want to test but not necessarily allow users to update to unless specifically instructed to. Non-snapshot version here should be fine IMO.

ptiurin (approved):

PR #237, opened 2023-06-28
Fixed unit tests that failed on Linux

alexradzin:
Hopefully the build became cross-platform (that should be verified later)


src/test/java/com/firebolt/jdbc/client/UsageTrackerUtilTest.java
```diff
@@ -185,35 +160,16 @@ void shouldParseDetectStack() {
...
-			assertEquals("ConnA/1.2.0 ConnB/3.0.4 JDBC/1.0-TEST (Java 8.0.1; Darwin 10.1; )", result);
-		}
+	@ParameterizedTest(name = "{1}")
+	@CsvSource(value = {
+			"98,less than 100,ConnA/1.1.1 ConnB/2.0.1",
+			"99,equals to 100,ConnA/1.1.1 ConnB/2.0.1",
+			"100, greater than 100,ConnA/1.2.0 ConnB/3.0.4",
+			"101, seriously greater than 100,ConnA/1.2.0 ConnB/3.0.4"
+	}, delimiter = ',')
+	void shouldNotOverrideConnectorsIfMoreThanSeveralNameVersionPairsAreSpecified(int n, String name, String expectedUserAgentPrefix) {
```
ptiurin, line 170: Is `name` used anywhere or is it added for readability?

ptiurin (approved):
Apart from that, code LGTM. 
Since you're changing the target compatibility you'll need to check that it sits well with our connectors that are running JDBC. Also that Peacekeeper does not run on java 8 or something.

stepansergeevitch (approved):
LGTM. As mentioned, let's not release this until we have a successful peacekeeper run

src/test/java/com/firebolt/jdbc/client/UsageTrackerUtilTest.java
alexradzin: yes, it is added for readability for similarity with the previous implementation where each such test was implemented in  separate function. "less than 100" is clearer than 99

PR #239, opened 2023-07-10
Bug/fir 24712 time functions

alexradzin:
fixed almost all problems of timestamp support. One test is partially disabled for now.

### DIFF
--- a/src/integrationTest/java/integration/tests/TimestampTest.java
+++ b/src/integrationTest/java/integration/tests/TimestampTest.java
@@ -96,14 +96,13 @@ class TimestampTest extends IntegrationTest {
 
 	@Test
 	@DefaultTimeZone("Asia/Kolkata")
-	@Disabled("FIR-24712")
 	void shouldRemoveOffsetDIffWhenTimestampOffsetHasChanged() throws SQLException {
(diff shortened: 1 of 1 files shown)


ptiurin (approved):
lgtm, let's file a ticket on that commented out test so we don't forget to fix it.

src/integrationTest/java/integration/tests/TimestampTest.java
```diff
@@ -279,7 +278,7 @@ private void compareAllDateTimeResultSetValuesWithPostgres(ResultSet fireboltRes
 		try (Connection postgresConnection = embeddedPostgres.getPostgresDatabase().getConnection();
 				Statement pgStatement = postgresConnection.createStatement()) {
 			if (timezone != null) {
-				pgStatement.execute(String.format("set timezone = '%s'", timezone));
```
stepansergeevitch, line 282: Please explain why did we have to remove the equals sign?

stepansergeevitch (approved):
Generally LGTM, one single comment before merging

src/integrationTest/java/integration/tests/TimestampTest.java
alexradzin: Good catch. Fixed.

PR #245, opened 2023-07-19
ci: new identity

stepansergeevitch:
Update integration testing workflow to use a new setup action v2, which creates new identity engines and databases

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -8,8 +8,8 @@ on:
         required: false
         default: ''
       engine:
-        description: 'Engine'
-        required: true
+        description: 'Engine - a new one will be created if not provided'
+        required: false
       account:
         description: 'Account'
         required: true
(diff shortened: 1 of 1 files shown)


alexradzin (approved):

.github/workflows/integration-test.yml
```diff
@@ -8,12 +8,12 @@ on:
         required: false
         default: ''
       engine:
-        description: 'Engine'
-        required: true
+        description: 'Engine - a new one will be created if not provided'
+        required: false
```
ptiurin, line 12: This is  somewhat misleading, if database is provided but engine is blank we'll break on `find-engine-name` step. Should we check that both inputs are either defined or undefined?

stepansergeevitch: Makes sense, will add this

.github/workflows/integration-test.yml
```diff
@@ -42,25 +42,29 @@ jobs:
...
+             echo "SERVICE_ACCOUNT_ID=${{ secrets.FIREBOLT_CLIENT_ID_NEW_IDN }}" >> "$GITHUB_ENV"
+             echo "SERVICE_ACCOUNT_SECRET=${{ secrets.FIREBOLT_CLIENT_SECRET_NEW_IDN }}" >> "$GITHUB_ENV"
           fi
+
+      - name: Validate database and engine
+        if: ${{ (github.event.inputs.database == '') != (github.event.inputs.engine == '') }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+            core.setFailed("Database and Engine parameters should be provided simultaneously")
```
ptiurin, line 57: Nitpicking, but let's set this as a first step in a workflow. No point in checking out the code and setting up java if we were to fail anyway. Saves some time.

ptiurin (approved):
Apart from that lgtm.

PR #248, opened 2023-07-21
VERSION 2.4.5-SNAPSHOT

alexradzin:


ptiurin:
Should https://github.com/firebolt-db/jdbc/blob/master/build.gradle#L12 be changed instead? Also your build is failing on a version check I believe.

ptiurin (approved):
lgtm

PR #251, opened 2023-07-23
FIR-25184 move driver version to one place - gradle.properties - and …

alexradzin:
…expose some properties to MANIFEST.MF


build.gradle
```diff
@@ -216,6 +215,24 @@ publishing {
     }
 }
 
+jar {
+    manifest {
+        attributes(
+                'Implementation-Title': 'Firebolt JDBC driver', // This value must be equal to one defined in VersionUtil
+                'Implementation-Vendor': 'Firebolt Analytics',
+                'Implementation-Version': version,
```
stepansergeevitch, line 223: Will this work accounting that you've removed `version` variable?

alexradzin: It is not removed but rather moved to gradle.properties

stepansergeevitch (approved):

PR #253, opened 2023-07-24
More tests

alexradzin:


src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
```diff
@@ -313,12 +313,12 @@ public int getDatabaseMinorVersion() throws SQLException {
 
 	@Override
 	public int getJDBCMajorVersion() throws SQLException {
-		return VersionUtil.getMajorDriverVersion();
+		return 4;
```
stepansergeevitch, line 316: Why are these values hardcoded?

src/test/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadataTest.java
```diff
@@ -227,6 +298,31 @@ void shouldGetTables() throws SQLException {
 		AssertionUtil.assertResultSetEquality(expectedResultSet, resultSet);
 	}
 
+	@Test
+	void shouldGetDriverMajorVersion() {
+		assertEquals(2, fireboltDatabaseMetadata.getDriverMajorVersion());
+	}
+
+	@Test
+	void shouldGetDriverManorVersion() {
```
stepansergeevitch, line 307: nit: Manor->Minor

src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
alexradzin: It is implemented in the second branch. I will merge it first, rebase this one and it will become not hard coded.

alexradzin: I've rebased this branch on master and fixed this implementation: now it is not hard coded but the value is read from gradle.properties

stepansergeevitch (approved):

PR #254, opened 2023-07-31
FIR-25300 improved discovery of isReadOnly() for connection and datab…

alexradzin:
…ase metadata.

### DIFF
--- a/src/integrationTest/java/integration/tests/IsReadableTest.java
+++ b/src/integrationTest/java/integration/tests/IsReadableTest.java
@@ -1,0 +1,42 @@
+package integration.tests;
+
+import integration.IntegrationTest;
+import lombok.CustomLog;
+import org.junit.jupiter.api.Test;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@CustomLog
+public class IsReadableTest extends IntegrationTest {
+    @Test
+    void connection() throws SQLException {
+        try (Connection connection = createConnection()) {
+            assertFalse(connection.isReadOnly());
+        }
+    }
+
+    @Test
+    void databaseMetadata() throws SQLException {
+        try (Connection connection = createConnection()) {
+            assertFalse(connection.getMetaData().isReadOnly());
+        }
+    }
+
+    @Test
+    void systemConnection() throws SQLException {
+        try (Connection connection = createConnection(getSystemEngineName())) {
+            assertTrue(connection.isReadOnly());
+        }
+    }
+
+    @Test
+    void systemDatabaseMetadata() throws SQLException {
+        try (Connection connection = createConnection(getSystemEngineName())) {
+            assertTrue(connection.getMetaData().isReadOnly());
+        }
+    }
+}
--- a/src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
+++ b/src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
@@ -9,8 +9,15 @@ import lombok.NonNull;
 import lombok.Value;
 import lombok.experimental.UtilityClass;
 
+import static java.lang.String.format;
+
 @UtilityClass
 public class MetadataUtil {
+	// lower() is used in SQL statement because the comparison in Firebolt is case-sensitive but we retrieve engine name from host where lowercase is used.
+	// But default engine is created with suffix Ingest (upper I). This is the reason to call lower() function.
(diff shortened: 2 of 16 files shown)


src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
```diff
@@ -102,4 +102,14 @@ public static String getDatabaseVersionQuery(String engine) {
 				.conditions(Collections.singletonList(String.format("engine_name iLIKE '%s%%'", engine))).build()
 				.toSql();
 	}
+
+	// lower() is used in SQL statement because the comparison in Firebolt is case-sensitive but we retrieve engine name from host where lowercase is used.
+	// But default engine is created with suffix Ingest (upper I). This is the reason to call loser() function.
```
ptiurin, line 107: typo: `lower`

src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
```diff
@@ -102,4 +102,14 @@ public static String getDatabaseVersionQuery(String engine) {
...
+	public static String isConnectionReadOnly() {
+		return "select count(*) = 0 from information_schema.engines where status = 'Running' and engine_type = 'General Purpose' and lower(engine_name) = lower(?)";
+	}
+
+	public static String isDatabaseReadOnly() {
+		return "select count(*) = 0 from information_schema.engines where status = 'Running' and engine_type = 'General Purpose' and lower(attached_to) = lower(?)";
+	}
```
ptiurin, line 114: This is fine for Firebolt 1.0, but just a heads-up - FB 2.0 will have RBAC so cases when a user doesn't have access to GP engine or user is set to reader despite working on a GP engine will be possible.

src/integrationTest/java/integration/tests/IsReadableTest.java
```diff
@@ -0,0 +1,68 @@
...
+            log.info("end of the engines list");
+
+            PreparedStatement ps2 = connection.prepareStatement("select count(*) = 0 from information_schema.engines where status = 'Running' and engine_type = 'General Purpose' and lower(attached_to) = lower(?)");
+            ps2.setString(1, database);
+            ResultSet rs2 = ps2.executeQuery();
+            while (rs2.next()) {
+                boolean isReadOnly = rs2.getBoolean(1);
+                log.info("read only={}", isReadOnly);
+            }
+            log.info("end of the readonly");
```
ptiurin, line 41: Why are we doing those statements in a test? Is this leftover from debugging?

src/test/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadataTest.java
```diff
@@ -355,7 +356,35 @@ void shouldGetDatabaseMinorVersion() throws SQLException {
...
+//		when(fireboltConnection.prepareStatement(anyString())).thenReturn(ps);
+//		when(ps.executeQuery()).thenReturn(rs);
+//		when(rs.next()).thenReturn(false);
+//		assertFalse(fireboltDatabaseMetadata.isReadOnly());
+
+
+//		ResultSet rs  = mock(ResultSet.class);
+//		when(rs.next()).thenReturn(false);
+//		isReadOnly(rs, false);
+	}
```
ptiurin, line 370: comments

src/integrationTest/java/integration/tests/IsReadableTest.java
alexradzin: no, this code is added for debugging.

src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
alexradzin: 10x

src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
alexradzin: Well... Do you have better solution? I will appreciate any help.

src/integrationTest/java/integration/tests/IsReadableTest.java
alexradzin: removed redundant code

src/main/java/com/firebolt/jdbc/metadata/MetadataUtil.java
ptiurin: Don't know. You'll need to check with the relevant team if there's a way to verify access. #sql-frontent will probably be a good place to start. This is the only driver that exposes `is_ReadOnly` so we haven't had the same issue elsewhere.
Now that V1 and V2 are merged `is_ReadOnly` will potentially return false info with RBAC so I'm not sure if we should implement it in the current state.

PR #258, opened 2023-08-09
Access token can be now supplied using connection properties

alexradzin:


src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -124,6 +124,11 @@ public Optional<String> getAccessToken() throws FireboltException {
 	}
 
 	private Optional<String> getAccessToken(FireboltProperties fireboltProperties) throws FireboltException {
+		String accessToken = fireboltProperties.getAccessToken();
```
stepansergeevitch, line 127: As I understand, currently if access_token is provided we ignore username and password. Can we have a validation that only one of `access_token` and `username/password` is provided? If both are, throw an error (or since it's internal usage, maybe a warning log would be enough, up to you)

alexradzin: done: I prefer to throw exception. Test is added too.

stepansergeevitch (approved):

PR #261, opened 2023-08-13
removed redundant and updated versions of some dependencies

alexradzin:
This reduces size of resulting jar by ~30% (from 11MB to 7MB)


stepansergeevitch (approved):

src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
```diff
@@ -101,11 +98,9 @@ private static String getArraySeparator(int dimension) {
 		return stringBuilder.toString();
 	}
 
+	@SuppressWarnings("java:S5852") // simplicity of code that uses regex here is more important than pre-mature optimisation
 	private static String removeQuotesAndTransformNull(String s) {
-		if (StringUtils.equals(s, "NULL")) {
-			return "\\N";
-		}
-		return CharMatcher.is('\'').trimFrom(s);
+		return "NULL".equals(s) ? "\\N" : s.replaceFirst("^'+", "").replaceFirst("'+$", "");
```
ptiurin, line 103: I believe `S5852` is concerned about the DOS vulnerability of this regex rather than optimisation. Are we safe here?

alexradzin: @ptiurin, you are right. I was astonished as well. The explanation here is that regex generally can cause performance issues. However in fact I do not think that this is real problem here.

src/main/java/com/firebolt/jdbc/util/StringUtil.java
ptiurin, line 1: Are we able to use `strip` from Apache Commons? It's not using a regex so should be safe.

https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#strip(java.lang.String,%20java.lang.String)

alexradzin: Done. I'm embarrassed to admit that I did not know this function. Thank you.

ptiurin (approved):
lgtm

PR #272, opened 2023-09-05
More tests2

alexradzin:


src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -514,87 +528,75 @@ public void setHoldability(int holdability) throws SQLException {
...
 
 	@Override
-	@ExcludeFromJacocoGeneratedReport
 	@NotImplemented
 	public NClob createNClob() throws SQLException {
 		throw new SQLFeatureNotSupportedException();
 	}
 
 	@Override
-	@ExcludeFromJacocoGeneratedReport
```
ptiurin, line 577: Can you expand on what is this doing and why is it being removed?

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -433,8 +430,6 @@ public void setFetchDirection(int direction) throws SQLException {
 	}
 
 	@Override
-	@NotImplemented
-	@ExcludeFromJacocoGeneratedReport
 	public int getFetchSize() throws SQLException {
 		return 0;
 	}
```
ptiurin, line 435: This should stay "not implemented". From Oracle's docs, this seems to be the pagination size so 0 would be misleading. https://docs.oracle.com/cd/E11882_01/java.112/e16548/resltset.htm#JJDBC28621

src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
ptiurin, line 1: Why are we removing `	@NotImplemented` in this file? Those methods seem to be not implemented.

stepansergeevitch (approved):
Always good to have more tests

ptiurin (commented):
It's somewhat misleading that PR title is "more tests", since there are also changes to functionality.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
alexradzin: This annotation excludes the code of the method from the test coverage report. I have added test that validates this method, so the annotation can be removed.

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
alexradzin: restored `@NotImplemented`

ptiurin (approved):

PR #280, opened 2023-09-28
FIR-26736 improved support of infinity

alexradzin:
Also added conversion from int floating point types

### DIFF
--- a/src/main/java/com/firebolt/jdbc/type/BaseType.java
+++ b/src/main/java/com/firebolt/jdbc/type/BaseType.java
@@ -24,10 +24,10 @@ import java.util.TimeZone;
 /** This class contains the java types the Firebolt datatypes are mapped to */
 @CustomLog
 public enum BaseType {
-	LONG(Long.class, conversion -> Long.parseLong(conversion.getValue())),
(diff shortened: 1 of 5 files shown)


src/main/java/com/firebolt/jdbc/type/BaseType.java
```diff
@@ -147,6 +147,13 @@ private <T> T transform(StringToColumnTypeConversion conversion) throws SQLExcep
 		return (T) transformFunction.apply(conversion);
 	}
 
+	private static String checkInfinity(String s) {
+		if ("inf".equals(s) || "-inf".equals(s)) {
```
ptiurin, line 151: Let's create const/enum (whichever is java's best practice) for those inf values. This way it will be easier to use them elsewhere if needed.

ptiurin (approved):
Apart from that lgtm

src/main/java/com/firebolt/jdbc/type/BaseType.java
alexradzin: Good comment, but I did something else: reused already existing functions

PR #283, opened 2023-10-04
publishing artifact to the central maven repository

alexradzin:


stepansergeevitch (approved):
LGTM

build.gradle
```diff
@@ -244,6 +299,11 @@ task generateJavadoc(type: Javadoc) {
     options.addStringOption('Xdoclint:none', '-quiet')
 }
 
+
+signing {
+    sign publishing.publications
+}
```
ptiurin, line 305: How does signing work on github? Do we store the private key somewhere?

ptiurin (commented):
Otherwise lgtm. However, I'm thinking should we publish to both repsy and MC? Since we're trying to keep 2.* version of JDBC backwards compatible. Otherwise any existing users won't get any future updates to 2.* JDBC. We can remove repsy altogether for JDBC 3.* since that one is a major version.

build.gradle
alexradzin: yes, I have added a secret. I am planning write document on Sunday

ptiurin:
> Generally it is possible, but I do not understand why. Maven central + sonatype a standard, "default" repositories. Artefacts published there are available for any build that references it as a dependency without mentioning the repository URL. 

Would users that had previously used repsy and still reference repsy in their maven config get updates to JDBC 2.*? I don't understand why are we making those changes without considering the impact.

> I even do not know why Repsy has been chosen

https://packboard.atlassian.net/wiki/spaces/INTERFACES/pages/2887057554/JDBC+connector+deployment

PR #306, opened 2023-12-06
Fix for string escape

dusanstanojeviccs:
Fixes an escape issue for single quotes when using PreparedStatement.


dusanstanojeviccs:
Screenshot 2023-12-07 at 11 08 30 AM

Closing this as it seems that there is a flag for fixing this that the Firebolt team can enable while they figure out the code transition path

PR #307, opened 2023-12-12
Fir 28437 use

alexradzin:


src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -16,14 +16,19 @@
...
 
 import static java.lang.String.format;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class ConnectionTest extends IntegrationTest {
+    private static final String DATABASE_NAME = "jdbc_use_db_integration_test";
```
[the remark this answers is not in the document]
alexradzin: This irrelevant here anyway, so I have removed these constants. The comment however is relevant for `SystemEngineTest`.

src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -80,6 +86,15 @@ void ddl() throws SQLException {
 		}
 	}
 
+	@Test
+	@Tag("dev")
```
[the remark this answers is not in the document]
alexradzin: I've created separate test that should fail on staging.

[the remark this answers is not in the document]
ptiurin: I see. It still feels a bit awkward to pass excludeTags=staging each time. Can this be inferred by the tests somehow? i.e. we know that env for staging contains "staging". Maybe the test can check this and "skip itself" if correct environment is not detected?
I'm just thinking a number of parameters we need to supply to successfully run tests locally keeps on growing.

src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -33,10 +35,15 @@
 public class SystemEngineTest extends IntegrationTest {
 
 	private static final long ID = ProcessHandle.current().pid() + System.currentTimeMillis();
-	private static final String DATABASE_NAME = "jdbc_system_engine_integration_test_" + ID;
-	private static final String ENGINE_NAME = DATABASE_NAME + "_engine";
+	private static final String SECOND_DATABASE_NAME = "jdbc_system_engine_integration_test_" + ID;
+	private static final String ENGINE_NAME = SECOND_DATABASE_NAME + "_engine";
 	private static final String ENGINE_NEW_NAME = ENGINE_NAME + "_2";
 
+	private static final String USE_DATABASE_NAME = "jdbc_use_db_" + ID;
```
ptiurin, line 42: Now this is outside of scope, but why do we generate the ID here? We already have a unique database name provided to the test suite in `-Ddb=`. We can just do that db name + "system_engine". At least this way we'll be able to match system engine test dbs to a particular JDBC test run and keep test db prefixes consistent.

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -284,6 +285,19 @@ private Map<String, String> getCancelParameters(String statementId) {
...
 
+
+	@Override
+	protected void validateResponse(String host, Response response, Boolean isCompress) throws FireboltException {
+		super.validateResponse(host, response, isCompress);
+		FireboltConnection connection = getConnection();
+		if (isCallSuccessful(response.code())) {
+			for (String header : response.headers(HEADER_UPDATE_PARAMETER)) {
+				String[] keyValue = header.split("=");
+				connection.addProperty(Pair.of(keyValue[0].trim(), keyValue[1].trim()));
```
stepansergeevitch, line 296: I think we should only add known parameters from this header (as we do in python).
Just have a list of parameters that we know (basically only "database") and add a property only in case key is in this list

ptiurin (approved):
Couple of suggestions, but otherwise LGTM.
Also please change the PR name to something more descriptive before merging. e.g. `FIR-28437 support use command` or implement use command. IMO, it's easier to understand what the change was about without going into it.

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionUserPassword.java
```diff
@@ -33,14 +33,14 @@ public class FireboltConnectionUserPassword extends FireboltConnection {
...
-        super(url, connectionSettings, fireboltAuthenticationService, fireboltStatementService);
+        super(url, connectionSettings, fireboltAuthenticationService, fireboltStatementService, "1.0");
         this.fireboltEngineService = fireboltEngineService;
         connect();
     }
 
     @ExcludeFromJacocoGeneratedReport
     FireboltConnectionUserPassword(@NonNull String url, Properties connectionSettings) throws SQLException {
-        super(url, connectionSettings);
+        super(url, connectionSettings, "1.0");
```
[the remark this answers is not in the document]
stepansergeevitch: Let's use null. This way we have less possible cases for backend and are consistent with older versions

src/integrationTest/java/integration/tests/SystemEngineTest.java
alexradzin: The ID is needed here because If test failed or interrupted for some reason the DB is still there and the next execution of the test fails. I saw it several times when running test locally. Also clash can happen when running intetegration tests twice. This happened to me too: I ran tests, fixed someting and ran them again. One of the executions failed because other one created DB first.

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
alexradzin: Definitely not. This will work for `set` statement too.

src/integrationTest/java/integration/tests/SystemEngineTest.java
ptiurin: Then those tests are setup incorrectly, we should look into that. `integration-setup-action` on each run provides a unique db & engine name. Otherwise it's a hack that's there because some tests are not using the db name provided.

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
stepansergeevitch: We don't yet have `set` statements and we don't know yet how they will be implemented. The header approach had some issues so it was postponed. I believe we should enable this functionality excplicitly to make sure it's tested before it's exposed to customers and not just leave backdoors for it to work in the future

PR #311, opened 2023-12-21
FIR-28894: Allow looking up engine URIs and database schemas in catalog 

maximilian-reinsel:
What:
- Change queries we use to look up the engine URI and default database from an engine to account for catalog information_schema table.
- Pass along any query params that come with the uri that we look up


alexradzin:
Alternative implementation is done.

PR #313, opened 2023-12-25
FIR-28894 support of v2 engines

alexradzin:


stepansergeevitch (approved):

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -66,12 +81,13 @@ private Engine getEngine(String engine, @Nullable String database) throws SQLExc
                 if (database != null && !database.equals(attachedDatabase)) {
                     throw new FireboltException(format("The engine with the name %s is not attached to database %s", engine, database));
                 }
-                return new Engine(rs.getString(ENGINE_URL), status, rs.getString(ENGINE_NAME_FIELD), attachedDatabase, null);
+                String url = rs.getString(ENGINE_URL).split("\\?", 2)[0];
```
maximilian-reinsel, line 84: In https://github.com/firebolt-db/jdbc/pull/311 i was grabbing the query params because we need to set those to get correct routing to the engine if there are any present.

maximilian-reinsel (changes requested):
We need to set the query params that come back in the url

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -11,29 +11,44 @@
...
 public class FireboltEngineInformationSchemaService implements FireboltEngineService {
     private static final String ENGINE_URL = "url";
     private static final String STATUS_FIELD = "status";
     private static final String ENGINE_NAME_FIELD = "engine_name";
-    private static final String RUNNING_STATUS = "running";
+    private static final Collection<String> RUNNING_STATUSES = Stream.of("running", "ENGINE_STATE_RUNNING").collect(toCollection(() -> new TreeSet<>(CASE_INSENSITIVE_ORDER)));
     private static final String ENGINE_QUERY =
             "SELECT engs.url, engs.attached_to, dbs.database_name, engs.status, engs.engine_name " +
             "FROM information_schema.engines as engs " +
             "LEFT JOIN information_schema.databases as dbs ON engs.attached_to = dbs.database_name " +
```
maximilian-reinsel, line 32: We need to look at catalogs here potentially

maximilian-reinsel: nevermind, the information_schema.databases table works now here.

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -11,29 +11,44 @@
...
 @RequiredArgsConstructor
 @CustomLog
 public class FireboltEngineInformationSchemaService implements FireboltEngineService {
     private static final String ENGINE_URL = "url";
     private static final String STATUS_FIELD = "status";
     private static final String ENGINE_NAME_FIELD = "engine_name";
-    private static final String RUNNING_STATUS = "running";
+    private static final Collection<String> RUNNING_STATUSES = Stream.of("running", "ENGINE_STATE_RUNNING").collect(toCollection(() -> new TreeSet<>(CASE_INSENSITIVE_ORDER)));
     private static final String ENGINE_QUERY =
             "SELECT engs.url, engs.attached_to, dbs.database_name, engs.status, engs.engine_name " +
```
maximilian-reinsel, line 30: In v2, engs.attached_to will always be empty. We could consider just removing it from this query + our check if it is empty or not - see https://firebolt-analytics.slack.com/archives/C02QFG24RN1/p1704838365433459?thread_ts=1702678581.390369&cid=C02QFG24RN1 for some details.

maximilian-reinsel: The check below 
```
                String attachedDatabase = rs.getString("attached_to");
                if (attachedDatabase == null) {
                    throw new FireboltException(format("The engine with the name %s is not attached to any database", engine));
                }
```

will still fail for any v2 engine.

maximilian-reinsel (changes requested):

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
alexradzin: Done.

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -62,7 +62,9 @@ public class StatementClientImpl extends FireboltClient implements StatementClie
 	//private static final String QUERY_ID_FETCHER = "select query_id from information_schema.query_history where status = 'STARTED_EXECUTION' and  query_label = ?";
 	private static final String QUERY_ID_FETCHER = "select query_id from information_schema.query_history where status = 'STARTED_EXECUTION' and  query_text like ?";
 	private final BiPredicate<Call, String> isCallWithLabel = (call, label) -> call.request().tag() instanceof String && Objects.equals(call.request().tag(), label);
-	static final String HEADER_UPDATE_PARAMETER = "Firebolt-Update-Parameters";
+	public static final String HEADER_UPDATE_PARAMETER = "Firebolt-Update-Parameters";
+	public static final String HEADER_UPDATE_ENDPOINT = "Firebolt-Update-Endpoint";
+	public static final String HEADER_RESET_SESSION = "Firebolt-Update-Session";
```
ptiurin, line 67: typo, `Firebolt-Reset-Session` header according to the spec

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -290,6 +292,12 @@ protected void validateResponse(String host, Response response, Boolean isCompre
 		super.validateResponse(host, response, isCompress);
 		FireboltConnection connection = getConnection();
 		if (isCallSuccessful(response.code())) {
+			if (response.header(HEADER_UPDATE_ENDPOINT) != null) {
+				connection.setEndpoint(response.header(HEADER_UPDATE_ENDPOINT));
+			}
```
ptiurin, line 297: Endpoint returned from the backend would contain query parameters e.g. `http://url.io/?account_id=1&database_name=b`. We want to parse it, do error checks (e.g. returned account id is the same that we hold) and I'd assume keep only the url part. Please verify with the design doc.

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -66,12 +79,16 @@ private Engine getEngine(String engine, @Nullable String database) throws SQLExc
                 if (database != null && !database.equals(attachedDatabase)) {
                     throw new FireboltException(format("The engine with the name %s is not attached to database %s", engine, database));
                 }
-                return new Engine(rs.getString(ENGINE_URL), status, rs.getString(ENGINE_NAME_FIELD), attachedDatabase, null);
+                String[] engineUrl = rs.getString(ENGINE_URL).split("\\?", 2);
+                String engineHost = engineUrl[0];
+                String[] engineQuery = engineUrl.length > 1 ? engineUrl[1].split("&") : new String[0];
+                Arrays.stream(engineQuery).map(prop -> prop.split("=")).filter(a -> a.length == 2).forEach(prop -> properties.addProperty(prop[0], prop[1]));
```
ptiurin, line 85: Let's pop a comment here what's happening, otherwise a long string of map/reduce is hard to read.

ptiurin (commented):
Other notes,

- `PROTOCOL_VERSION` has to be bumped to 2.1
- Based on account version we want to disable some of the checks (e.g. that database is connected to an engine) and use "USE DATABASE/ENGINE" when connecting.
- Let's add Integration tests

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
alexradzin: fixed

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
alexradzin: I do not see this in spec:

PackDB appends the following HTTP headers to the response:
Firebolt-Update-Endpoint: https://new-endpoint.example.com/
Firebolt-Update-Parameters: engine=engine_name,account_id=account_id
Firebolt-Reset-Session:
client updates its map of parameters

ptiurin: As discussed on the daily that's PackDB design doc, let's refer to the implementation doc that was created for Ecosystem specifically https://docs.google.com/document/d/1f0kdPCPbhCrmMTK-sfqAkqEkonT7xyNK-teJmXE_6p0/

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
alexradzin: Fixed. Now we have 3 implementations of `FireboltEngineService`; one of them `FireboltEngineVersion2Service` implements this logic

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
alexradzin: done

ptiurin (commented):
Some questions and suggestions.
Also, let's get that 4% more coverage for Sonar to pass! 💪

.github/workflows/integration-test.yml
```diff
@@ -55,7 +66,18 @@ jobs:
...
+      FIREBOLT_CLIENT_ID_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}
+      FIREBOLT_CLIENT_SECRET_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}
+
+  run-integration-tests-engine2:
+    if: ${{ github.event.inputs.run-real-v2 == 'true' }}
+    uses: ./.github/workflows/integration-test-v2.yml
+    with:
+      database: ${{ github.event.inputs.database2real }}
+      engine: ${{ github.event.inputs.engine2real }}
+      account: "infra-engines-v2"
```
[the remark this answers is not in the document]
alexradzin: somehow it did not work, but I restore this.

src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -21,39 +23,60 @@
...
     enum EngineType {
         SYSTEM_ENGINE, CUSTOM_ENGINE
     }
 
+    @BeforeAll
+    void beforeAll() throws SQLException {
+        try (Connection conn = createConnection()) {
+            infraVersion = ((FireboltConnection) conn).getInfraVersion();
+        }
+    }
```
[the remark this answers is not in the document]
alexradzin: This is much simpler. Otherwise I need v2 and "realv2". I really spent *days* trying to find better solution and this is the best one I can find.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
```diff
@@ -236,6 +236,12 @@ public void addProperty(@NonNull String key, String value) {
 		switch (key) {
 			case "database": database = value; break;
 			case "engine": engine = value; break;
+			case "account_id":
+				if (accountId != null && !accountId.equalsIgnoreCase(value)) {
+					throw new IllegalStateException(format("Attempt to change account from %s to %s", accountId, value));
+				}
+				this.accountId = value;
```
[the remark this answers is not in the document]
alexradzin: Where? I do no see such code. IMHO the only place where we check that the "new" account is the same as the "old" one is here.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
```diff
@@ -236,6 +236,12 @@ public void addProperty(@NonNull String key, String value) {
 		switch (key) {
 			case "database": database = value; break;
 			case "engine": engine = value; break;
+			case "account_id":
+				if (accountId != null && !accountId.equalsIgnoreCase(value)) {
+					throw new IllegalStateException(format("Attempt to change account from %s to %s", accountId, value));
```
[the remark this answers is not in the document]
alexradzin: The message is changed although I really do not understand why suggested message is better than mine. Also I removed "USE" from the message because this method is called when processing response from any command; passing the executed command here requires a lot of changes that I do not want to do.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
```diff
@@ -249,4 +255,17 @@ public String getHttpConnectionUrl() {
...
 		return protocol + hostAndPort;
 	}
+
+	public void clearAdditionalProperties() {
+		additionalProperties.clear();
+	}
+
+	public String processEngineUrl(String endpoint) {
+		String[] engineUrl = endpoint.split("\\?", 2);
+		String engineHost = engineUrl[0];
```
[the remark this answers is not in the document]
alexradzin: Well, added, although it was never required.

src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -128,26 +125,56 @@ void shouldFailToSelectFromCustomDbUsingSystemEngine() throws SQLException {
...
+	void useEngine() throws SQLException {
+		try (Connection connection = createConnection(getSystemEngineName())) {
+			int infraVersion = ((FireboltConnection) connection).getInfraVersion();
+			assumeTrue(infraVersion >= 2);
+			try {
+				connection.createStatement().executeUpdate("USE ENGINE SYSTEM");
+				assertThrows(SQLException.class, () -> connection.createStatement().executeUpdate(format("USE ENGINE %s", ENGINE_NAME)));
+				connection.createStatement().executeUpdate(format("CREATE ENGINE %s", ENGINE_NAME));
+				connection.createStatement().executeUpdate(format("USE ENGINE %s", ENGINE_NAME));
+			} finally {
```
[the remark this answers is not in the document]
alexradzin: Good catch. Added and fixed a bug.

.github/workflows/integration-test.yml
```diff
@@ -55,7 +66,18 @@ jobs:
...
+    secrets:
+      FIREBOLT_CLIENT_ID_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}
+      FIREBOLT_CLIENT_SECRET_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}
+
+  run-integration-tests-engine2:
+    if: ${{ github.event.inputs.run-database-with-engine-v2 == 'true' }}
+    uses: ./.github/workflows/integration-test-v2.yml
+    with:
+      database: ${{ github.event.inputs.database-with-engine-v2 }}
+      engine: ${{ github.event.inputs.engine_v2_fb_2_0 }}
```
ptiurin, line 79: Let's take this opportunity to change all `github.event.inputs` to `inputs`. As we saw elsewhere this is a safer option.

.github/workflows/integration-test.yml
```diff
@@ -55,7 +66,18 @@ jobs:
...
+      account: ${{ vars. FIREBOLT_ACCOUNT_V2 }}
```
ptiurin, line 80: You have a space after `vars. `, let's remove it as I think this will fail.
This had a space in my previous comment so sorry about the confusion there.

src/integrationTest/java/integration/IfraVersionCondition.java
ptiurin, line 1: nit: typo in file/test name - `Ifra` -> `Infra`.
Just pointing out since searching for "infra" when trying to find this test later will show no results.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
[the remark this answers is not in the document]
ptiurin: Firebolt has a preferred error message format, which is explained in detail in [here](https://app.getguru.com/card/i6jo6ngT/Firebolt-Error-Message-Style-Guide). The messaged from the design doc is following this guideline.
Please remember to comment on the design doc if you have questions or suggestions, this helps us adjust before all the implementations are completed.

ptiurin (approved):
Minor comments, but please address them.
Otherwise looks good.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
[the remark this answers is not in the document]
alexradzin: Again: too  many changes to match randomly written document.

.github/workflows/integration-test.yml
alexradzin: Done, although making unrelated changes is the classical way to the hell.

.github/workflows/integration-test.yml
alexradzin: yes, fixed. I have seen it myself as well. Thank you.

src/integrationTest/java/integration/IfraVersionCondition.java
alexradzin: 10x,  fixed

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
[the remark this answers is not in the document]
alexradzin: I added my comment to the document. Obviously this comment is write only.

.github/workflows/integration-test.yml
ptiurin: I'd argue since we're adding these lines and we know them to be working incorrectly in certain cases it's related to this PR. We can either change only these lines but then we'll have two different calls in the same file or change them altogether and have a issue that we don't have to come back to.
This is not a dangerous operation/code that changed so in my opinion it's worth the effort to fix on the spot.

src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java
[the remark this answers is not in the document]
ptiurin: This document was written before we started the implementation and we follow it to keep the implementation logic consistent across the connectors. You can avoid doing "random" changes by reading the design document beforehand.

PR #317, opened 2023-12-31
added integration test of unauthorized account with 2 patches done be…

alexradzin:
…cause of the server side bugs

### DIFF
--- a/src/integrationTest/java/integration/tests/SystemEngineTest.java
+++ b/src/integrationTest/java/integration/tests/SystemEngineTest.java
@@ -213,22 +213,7 @@ public class SystemEngineTest extends IntegrationTest {
 //				String clientId = genKeyRs.getString(2);
 //				String clientSecret = genKeyRs.getString(3);
 				// But response of this command is incorrect (FIR-28997), so we have to retrieve clientId and clientSecret using SELECT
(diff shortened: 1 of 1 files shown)


src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -189,6 +198,53 @@ void useFailure() throws SQLException {
...
+				con.setRequestProperty("authorization", "Bearer " + accessToken);
+				con.setDoOutput(true);
+				try (PrintStream ps = new PrintStream(con.getOutputStream())) {
+					ps.println(format("CALL fb_GENERATESERVICEACCOUNTKEY('%s')", serviceAccountName));
+				}
+				BufferedReader br = new BufferedReader(new InputStreamReader(con.getInputStream(), UTF_8));
+				String[] data = br.readLine().split("\\t");
+//				String clientId = data[1];
+				String clientSecret = data[2];
+				// end of patch against FIR-28997
```
ptiurin, line 232: Maybe let's put this hack in a separate function to keep the test more or less readable.

ptiurin (approved):
Just a suggestion, otherwise lgtm I guess. That's a second bug we find in this small piece of functionality...

PR #319, opened 2024-01-03
added more unit tests and implementation of more functions of ResultSet

alexradzin:


src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
```diff
@@ -562,45 +564,35 @@ public <T> T unwrap(Class<T> iface) throws SQLException {
...
+		return getBinaryStream(columnLabel);
 	}
 
 	@Override
-	@NotImplemented
-	@ExcludeFromJacocoGeneratedReport
 	public InputStream getUnicodeStream(String columnLabel) throws SQLException {
-		throw new FireboltSQLFeatureNotSupportedException();
+		return getBinaryStream(columnLabel);
 	}
```
ptiurin, line 591: In these methods Postgres ensures only ASCII/Unicode values are returned. I think we should do the same, otherwise the user might get a unicode value from `getAsciiStream`, which is just incorrect. https://github.com/pgjdbc/pgjdbc/blob/aa235e14c2213d1386100e62206006d83082d914/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L2859-L2894

alexradzin: Good catch. Done

ptiurin (approved):

PR #323, opened 2024-01-10
FIR-29316 fixed support of multidimensional arrays

alexradzin:


src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
```diff
@@ -13,67 +13,100 @@
...
-			else
-				break;
-		value = value.substring(dimensions, value.length() - dimensions);
+		int dimensions = getDimensions(value);
 		Object arr = createArray(value, dimensions, columnType);
-		return new FireboltArray(columnType.getArrayBaseColumnType().getDataType(), arr);
+		return arr == null ? null : new FireboltArray(columnType.getArrayBaseColumnType().getDataType(), arr);
+	}
+
+	private static int getDimensions(String value) {
```
ptiurin, line 30: Do we have access to the data type? Dimensions can be inferred much easier from the type than value, e.g. `Array(Array(Array(INT) NOT NULL) NULL)` can be parsed by just counting a number of `Array`'s

alexradzin: Sure, I thought about it but it is not here. This function parses data without knowing the datatype. It is not what I changed. I just changed (fixed) the implementation that did not work correctly.

ptiurin: Right, but in the caller function, `transformToSqlArray` you have access to `columnType`. You can pass that to `getDimensions` instead of value and make the function a one-liner instead.

alexradzin: 1. I did not have an aim to change existing functionality. I just fixed the bug. 
2. There are 2 different things: data parsing and data type identification. We can parse data identifying the type from data itself, identify the data type and compare both types. If they are not compatible we can throw exception. Anyway this is not an issue that should be fixed in this PR. There are a lot of things that I do not like in code of the JDBC driver and this is one of them but not the worst one.

alexradzin: Following our discussion I have changed the implementation of `getDimensions()`, so hopefully it is good enough. At least it is definitely much shorter and simpler.

PR #324, opened 2024-01-11
updated versions of slf4j: 2.0.11 and jackson: 2.16.1

alexradzin:


alexradzin:
jackson 2.16.1  causes jarShadow task to fail. We have to postpone update of Jackson until the plugin is fixed. Here is existing PR: https://github.com/johnrengelman/shadow/pull/876
Here is issue that reports this problem https://github.com/johnrengelman/shadow/issues/894

PR #327, opened 2024-01-14
removed dependency on jaxb-api

alexradzin:


src/test/java/com/firebolt/jdbc/util/ByteArrayUtilTest.java
ptiurin, line 1: Let's add a couple more tests to be sure of quality. I'd suggest adding a test that parses empty string, single hex digit and invalid hex as well.

ptiurin (commented):
Also, how much are we gaining here? how heavy is the jaxb-api package?

src/test/java/com/firebolt/jdbc/util/ByteArrayUtilTest.java
alexradzin: Good idea. Tests are added and code is fixed and improved. Tests rock!

alexradzin:
> Also, how much are we gaining here? how heavy is the jaxb-api package?

https://packboard.atlassian.net/browse/FIR-29486

ptiurin (approved):

PR #331, opened 2024-01-29
fixed some integration tests broken due to changes at the server side

alexradzin:


src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -144,56 +140,23 @@ void shouldFailToSelectFromCustomDbUsingSystemEngine() throws SQLException {
...
-				Assert.assertThrows(SQLException.class, () -> connection.createStatement().executeUpdate(format("USE %s", USE_DATABASE_NAME))); // DB does not exist
-				connection.createStatement().executeUpdate(format("CREATE DATABASE IF NOT EXISTS %s", USE_DATABASE_NAME)); // create DB
-				connection.createStatement().executeUpdate(format("USE %s", USE_DATABASE_NAME)); // Now this should succeed
-				connection.createStatement().executeUpdate(format("CREATE TABLE %s ( id LONG)", TABLE2)); // create table2 in other DB
-				assertNull(getTableDbName(connection, TABLE1)); // table1 does not exist here
-				assertEquals(USE_DATABASE_NAME, getTableDbName(connection, TABLE2)); // but table2 does exist
+				systemConnection.createStatement().executeUpdate(format("USE %s", current.getDatabase())); // use current DB; shouldn't have any effect
+				assertNull(getTableDbName(systemConnection, TABLE1)); // the table does not exist yet
+				userConnection.createStatement().executeUpdate(format("CREATE TABLE %s ( id LONG)", TABLE1)); // create table1 in current DB
+				assertEquals(current.getDatabase(), getTableDbName(userConnection, TABLE1)); // now table t1 exists
```
ptiurin, line 151: Should we check if t1 exists on a systemConnection instead? Otherwise we're just confirming that we've run CREATE TABLE somewhere.

alexradzin: I'd love to but the validation is done by execution `select * from information_schema.tables`  that is not available in system engine. Attempt to execute this statement there causes error "Schema does not exists: information_shema". 

This is exactly the problem I complained yesterday: the `use` statement that worked well in dev does not work correctly at staging now. It works in system engine only while table can be created using user engine. This test was much more valuable in its previous version that worked on dev.

ptiurin: I see, then let's add `Assert.assertThrows` for this `information_schema` select here so when it finally gets fixed in system engine we can improve this test.

ptiurin (approved):
otherwise looks good

src/integrationTest/java/integration/tests/SystemEngineTest.java
alexradzin: Good idea. Done.

PR #333, opened 2024-01-31
FIR-29922: added support of +-nan and tests of nan

alexradzin:


stepansergeevitch (commented):
Please also remove any division by zero queries

ptiurin (approved):

PR #334, opened 2024-01-31
FIR-29770: removing dev environment

alexradzin:


.github/workflows/integration-test-v1.yml
```diff
@@ -50,23 +45,17 @@ jobs:
...
-             echo "USERNAME=${{ secrets.FIREBOLT_USERNAME_STAGING }}" >> "$GITHUB_ENV"
-             echo "PASSWORD=${{ secrets.FIREBOLT_PASSWORD_STAGING }}" >> "$GITHUB_ENV"
-             echo "EXCLUDE_ENV=dev" >> "$GITHUB_ENV"
-          else
-             echo "USERNAME=${{ secrets.FIREBOLT_USERNAME_DEV }}" >> "$GITHUB_ENV"
-             echo "PASSWORD=${{ secrets.FIREBOLT_PASSWORD_DEV }}" >> "$GITHUB_ENV"
-             echo "EXCLUDE_ENV=staging" >> "$GITHUB_ENV"
-          fi
+           echo "USERNAME=${{ secrets.FIREBOLT_USERNAME_STAGING }}" >> "$GITHUB_ENV"
+           echo "PASSWORD=${{ secrets.FIREBOLT_PASSWORD_STAGING }}" >> "$GITHUB_ENV"
```
ptiurin, line 49: We need to remove this step altogether, as it breaks when special characters are encountered. You can pass the secrets directly e.g. https://github.com/firebolt-db/dbt-firebolt/pull/120/files#diff-29ce8407d15551c9c62184e207b6ceadebb261f7ffbfb75d7d67844ba33709abR33

.github/workflows/integration-test.yml
ptiurin, line 1: DEV secrets should also be removed in L51-52

src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -177,27 +166,6 @@ void useSuccess() throws SQLException {
...
-				Assert.assertThrows(SQLException.class, () -> connection.createStatement().executeUpdate(format("USE %s", USE_DATABASE_NAME))); // DB exists but use statement is unsupported anyway
-			} finally {
-				try (Statement statement = connection.createStatement()) {
-					statement.executeUpdate(format("DROP DATABASE %s", USE_DATABASE_NAME));
-				} catch (SQLException e) { // catch just in case to do our best to clean everything even if test has failed
-					log.warn("Cannot drop database ",  USE_DATABASE_NAME, e);
-				}
-			}
-		}
-	}
```
ptiurin, line 199: Isn't this removal and `useSuccess` supposed to be other way around? Since we're removing dev, dev tests now can be deleted and staging kept by default.

.github/workflows/integration-test.yml
ptiurin: Same in other files.

src/integrationTest/java/integration/tests/SystemEngineTest.java
alexradzin: This should be and it is removed in other PR because staging now supports use statmenet

.github/workflows/integration-test-v1.yml
alexradzin: Hopefully fixed. Please review again.

src/integrationTest/java/integration/tests/SystemEngineTest.java
ptiurin: I feel like this is just a matter of resolving merge conflicts correctly, but ok. Makes sense.

ptiurin (approved):

PR #336, opened 2024-02-04
Suppress no javadoc warning

alexradzin:
Running `./gradlew build` produces a bunch of warning that report about missing javadoc. The default (?) policy requires javadoc on each public class/method/enum entry that is annoying. Meaning of some elements is obvious. For example `connect()` perform connection and `FireboltSessionProperty.DATABASE` is a property that holds database name. 

We do have some javadocs. Yes, some docs are missing. I will add documentation from time to time. But right now I'd want to suppress warnings about missing javadocs because I can miss other, more valuable warnings because of "missing javadoc" warnings.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,6 +98,7 @@ dependencies {
     compileTestCommonJava.dependsOn processTestResources
     compileTestCommonJava.dependsOn processIntegrationTestResources
     compileIntegrationTestJava.dependsOn processIntegrationTestResources
+    javadoc.options.addStringOption('Xdoclint:none', '-quiet')
 }
 
 test {


ptiurin (changes requested):
Merge artefacts present. Please make sure to verify the change before pushing.

ptiurin (commented):
What's the reason for this change? Is it not better to add javadoc where possible?

alexradzin:
> p

When I run `./gradlew build` I get 100 warnings. Why only 100? Because it is the default limit. Actually it reports about each class, public method and enum value that does not have javadoc. There is no need to write javadoc like "This function performs connection to database" on method `connect()` because it kind of obvious. Javadoc helps to understand the functionality only if it is not obvious from the method/class name otherwise it is annoying. 

I even wanted to write javadocs for all public classes but then I understood that this pointless work will take several days I decided to suppress these warnings so far. However I am going to add javadocs in places where these docs are needed.

ptiurin:
> When I run `./gradlew build` I get 100 warnings.

Let's try to add the reasoning for a change in the PR description, it helps understanding what's going on without the need to back and forth. I agree having hundreds of warnings is not great and should be fixed.

> Javadoc helps to understand the functionality only if it is not obvious from the method/class name otherwise it is annoying.
> I even wanted to write javadocs for all public classes but then I understood that this pointless work will take several days I decided to suppress these warnings so far.

Let's discuss this on a daily, but I'd add a counterpoint - google's Java coding standards at a minimum require public classes to have javadoc, though they acknowledge that some may not be needed e.g. getFoo() does not need a javadoc saying that this method returns Foo. Perhaps we can suppress on case-by-case basis?

alexradzin:
> > When I run `./gradlew build` I get 100 warnings.
> 
> Let's try to add the reasoning for a change in the PR description, it helps understanding what's going on without the need to back and forth. I agree having hundreds of warnings is not great and should be fixed.
> 
> > Javadoc helps to understand the functionality only if it is not obvious from the method/class name otherwise it is annoying.
> > I even wanted to write javadocs for all public classes but then I understood that this pointless work will take several days I decided to suppress these warnings so far.
> 
> Let's discuss this on a daily, but I'd add a counterpoint - google's Java coding standards at a minimum require public classes to have javadoc, though they acknowledge that some may not be needed e.g. getFoo() does not need a javadoc saying that this method returns Foo. Perhaps we can suppress on case-by-case basis?

I added PR description

stepansergeevitch (approved):
Approved, created a JIRA to add javadocs where they're missing: https://packboard.atlassian.net/browse/FIR-30302

PR #346, opened 2024-02-15
added getting version using class VersionUtils

alexradzin:
FIR-30343


ptiurin:
@alexradzin can we get some context here or in the ticket, when is the version missing from the Manifest? What about `version.properties`? Is there a case when we'll fall all the way through to the defaults?

alexradzin:
> FIR-30343

added javadoc that explains everything.

PR #350, opened 2024-02-20
FIR-30506: fixed connection without DB

alexradzin:
also added tests that validates (unsuccessful) connection without account

### DIFF
--- a/src/integrationTest/java/integration/tests/ConnectionTest.java
+++ b/src/integrationTest/java/integration/tests/ConnectionTest.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
+import static org.junit.jupiter.params.ParameterizedTest.ARGUMENTS_WITH_NAMES_PLACEHOLDER;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class ConnectionTest extends IntegrationTest {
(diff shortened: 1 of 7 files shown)


src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -99,8 +99,8 @@ void connect(boolean useDatabase, boolean useEngine) throws SQLException {
         String engine = useEngine ? params.getEngine() : null;
         ConnectionInfo updated = new ConnectionInfo(params.getPrincipal(), params.getSecret(), params.getEnv(), database, params.getAccount(), engine, params.getApi());
         String url = updated.toJdbcUrl();
-        String query = "SELECT TOP 1 * FROM information_schema.tables";
-        boolean expectedSuccess = updated.getApi() != null || infraVersion == 2 ? useDatabase : useDatabase || useEngine;
+        String query = "SELECT version()";
+        boolean expectedSuccess = updated.getApi() == null || useDatabase;
```
ptiurin, line 103: Isn't this test supposed to work even if `useDatabase` is false? 
In general, let's avoid adding complicated if logic like we had in this test before. It only leads to confusion on what the test is testing.

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -64,6 +54,11 @@ void notExistingDb() throws SQLException {
         assertEquals("Database my_db does not exist", assertThrows(FireboltException.class, () -> createConnection("jdbc:firebolt:?env=dev&account=dev", connectionProperties)).getMessage());
     }
 
+    @Test
+    void noAccount() {
+        assertEquals("Cannot connect: account is missing", assertThrows(FireboltException.class, () -> createConnection("jdbc:firebolt:db", connectionProperties)).getMessage());
+    }
+
```
ptiurin, line 61: Just checking, this test is only V2, right?
Otherwise, in V1 account is optional.

ptiurin (commented):
Couple of clarifying questions, otherwise looks good.

src/integrationTest/java/integration/tests/ConnectionTest.java
alexradzin: Agree. Will be fixed

alexradzin: Fixed

ptiurin (approved):

PR #356, opened 2024-03-11
FIR-30652: implementation of the byte array support

alexradzin:


src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
```diff
@@ -221,4 +225,31 @@ private static Object[] toObjectArray(Object array) {
...
+		return BYTE_ARRAY_PREFIX + Stream.generate(buffer::get).limit(buffer.capacity()).map(Integer::toHexString).collect(joining(separator));
+	}
+
+	@SuppressWarnings("java:S1168") // we have to return null here
+	public static byte[] hexStringToByteArray(String str) {
+		if (str == null) {
+			return null;
+		}
+		if (!str.startsWith(BYTE_ARRAY_PREFIX)) {
+			return str.getBytes();
```
ptiurin, line 244: Reading up on `getBytes`, we probably should pass UTF-8 charset to it otherwise it uses a system-default one, which makes the decoding behaviour system-dependent.
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes-java.nio.charset.Charset-

src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
```diff
@@ -221,4 +225,31 @@ private static Object[] toObjectArray(Object array) {
...
+		}
+		char[] chars = str.substring(2).toCharArray();
+		byte[] bytes = new byte[chars.length / 2];
+		for (int i = 0; i < chars.length; i += 2) {
+			bytes[i / 2] = (byte) ((Character.digit(chars[i], 16) << 4) + Character.digit(chars[i + 1], 16));
+		}
```
ptiurin, line 250: Is there a way to use `getBytes` here as well? Perhaps stripping `BYTE_ARRAY_PREFIX` from every place and then calling it? This way we don't have to maintain this bitwise shifting logic.

src/test/java/com/firebolt/jdbc/type/array/SqlArrayUtilTest.java
```diff
@@ -118,6 +119,41 @@ void shouldTransformNullArrayToNull(String type) throws SQLException {
...
+	@ParameterizedTest
+	@ValueSource(booleans = {true, false})
+	void nullByteArrayToString(boolean separateEachByte) {
+		assertNull(SqlArrayUtil.byteArrayToHexString(null, separateEachByte));
+	}
+
+	@ParameterizedTest
+	@CsvSource({
+			"ABC,false,\\x414243",
+			"abc,true,\\x61\\x62\\x63"
```
ptiurin, line 131: Let's add some special cases like `\`, newline character and space.

src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
```diff
@@ -77,11 +89,15 @@ public Class<?> getSourceType() {
 	}
 
 	public String transform(Object object) throws FireboltException {
+		return transform(object, false);
+	}
+
+	public String transform(Object object, boolean flag) throws FireboltException {
```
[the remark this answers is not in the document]
alexradzin: Renamed although this is only the special case for byte array. In future we will have to pass any generic parameter here of other cases that are not still implemented in ResultSet. However it is indeed more readable now.

src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
alexradzin: Funny. I was thing about this while writing this code but than forgot. Fixed.

src/test/java/com/firebolt/jdbc/type/array/SqlArrayUtilTest.java
alexradzin: Done and even fixed bug. 10x.

src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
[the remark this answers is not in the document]
ptiurin: > Same in `transformAny`

This function still has `flag`.

Also, with `transform` I still don't understand the variable. It's a boolean parameter that governs anything depending on the transform used (some may not even support it). Can we create a different transform class instead?

src/test/java/com/firebolt/jdbc/type/array/SqlArrayUtilTest.java
ptiurin: Let's add slash character for completeness. e.g. `AB\C' just to make sure that slash is not lost in any way.

alexradzin: Added both slash `/` and back slash `\`

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -426,6 +429,14 @@ private void validateConnection(FireboltProperties fireboltProperties, boolean i
...
 	}
 
+	private void discoverByteArrayFormat(Statement s) {
+		try {
+			s.execute(format("SELECT %1$s42%1$s42", SqlArrayUtil.BYTE_ARRAY_PREFIX));
+		} catch (SQLException e) {
+			usePrefixForEachByte = false;
+		}
+	}
+
```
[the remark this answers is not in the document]
alexradzin: I used `E` prefix, so this code became not needed and is removed now.

src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -115,6 +115,7 @@ void successfulConnectUnsuccessfulSelectV1(boolean useDatabase, boolean useEngin
         String url = getJdbcUrl(params, useDatabase, useEngine);
         try (Connection connection = DriverManager.getConnection(url, params.getPrincipal(), params.getSecret());
              Statement statement = connection.createStatement()) {
+            statement.executeQuery("SELECT 1");
```
ptiurin, line 118: was this leftover from debugging?

ptiurin (approved):
apart from that lgtm

PR #358, opened 2024-03-14
fix: Add protocol to system engine

ptiurin:
System engine now comes without protocol in URL. We need to account for this in code.


ptiurin:
Thanks for adding the tests. I can't review my own PR but lgtm.

alexradzin (approved):

PR #369, opened 2024-03-28
ability to create DB and engine inside integration test for engine v2

alexradzin:


.github/workflows/integration-test-v2.yml
```diff
@@ -72,5 +77,14 @@ jobs:
...
+        run: |
+          if [ $runsetup == 'true' ]; then
+             echo "run_setup=false" >> $GITHUB_OUTPUT
+          else
+             echo "run_setup=true" >> $GITHUB_OUTPUT
+          fi
+
       - name: Run integration tests
-        run: ./gradlew integrationTest -Ddb=${{ steps.find-database-name.outputs.database_name }} -Denv=staging -Dclient_secret="${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}" -Dclient_id="${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}" -Daccount="${{ inputs.account }}" -Dengine="${{ steps.find-engine-name.outputs.engine_name }}" -DexcludeTags=v1
+        run: ./gradlew integrationTest -Ddb=${{ steps.find-database-name.outputs.database_name }} -Denv=staging -Dclient_secret="${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}" -Dclient_id="${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}" -Daccount="${{ inputs.account }}" -Dengine="${{ steps.find-engine-name.outputs.engine_name }}" -DexcludeTags=v1 -Drun_setup="${{ steps.setup-env.outputs.run_setup }}"
```
ptiurin, line 90: I don't think you need `setup-env` step, you can just reference the input `-Drun_setup="${{ inputs.run_setup }}"`

.github/workflows/integration-test-v2.yml
```diff
@@ -72,5 +77,14 @@ jobs:
              echo "engine_name=${{ steps.setup.outputs.engine_name }}" >> $GITHUB_OUTPUT
           fi
 
+      - name: Setup env by build
+        id: setup-env
+        run: |
+          if [ $runsetup == 'true' ]; then
```
ptiurin, line 83: Typo? `$runsetup` is not defined anywhere.

.github/workflows/integration-test-v2.yml
alexradzin: I inverse the flag value. `run_setup == !runsetup`.
It is because from the github action percpective runsetup=true means run setup here, in github action. From the `EnvironmentManager` point of view "run setup = true" means that it should run the setup. If setup is done by github action it is not done by the `EnviromnmentManager` and vice versa.

.github/workflows/integration-test-v2.yml
alexradzin: ups, 10x. I do not understand how I thought that it works correctly. Fixed

stepansergeevitch (approved):

PR #371, opened 2024-03-31
FIR-31838 getting account_id from the server and not sending it as a …

alexradzin:
…parameter for system engine

### DIFF
--- a/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
+++ b/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
@@ -8,6 +8,7 @@ import com.firebolt.jdbc.client.authentication.FireboltAuthenticationClient;
 import com.firebolt.jdbc.client.authentication.ServiceAccountAuthenticationRequest;
 import com.firebolt.jdbc.client.gateway.GatewayUrlResponse;
 import com.firebolt.jdbc.connection.settings.FireboltProperties;
+import com.firebolt.jdbc.connection.settings.FireboltQueryParameterKey;
 import com.firebolt.jdbc.exception.FireboltException;
 import com.firebolt.jdbc.service.FireboltAccountIdService;
 import com.firebolt.jdbc.service.FireboltAuthenticationService;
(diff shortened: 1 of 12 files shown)


src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
```diff
@@ -100,15 +106,24 @@ private FireboltProperties getSessionPropertiesForSystemEngine(String accessToke
         String systemEngineEndpoint = fireboltGatewayUrlService.getUrl(accessToken, accountName);
         FireboltAccount account = fireboltAccountIdService.getValue(accessToken, accountName);
         infraVersion = account.getInfraVersion();
+        URL systemEngienUrl = UrlUtil.createUrl(systemEngineEndpoint);
+        Map<String, String> systemEngineUrlUrlParams = UrlUtil.getQueryParameters(systemEngienUrl);
+        String accountId = systemEngineUrlUrlParams.getOrDefault(ACCOUNT_ID.getKey(), account.getId());
```
stepansergeevitch, line 111: I think the idea here is to use all parameters and not only `account_id`. Yes, currently (and maybe even further) we only receive `account_id`, but generally this assumes the ability to receive other parameters.
So generally, we should take all received parameters (not only `account_id`) and add them when we connect to a system engine.

stepansergeevitch: Also, if account_id is not provided in parameters, we should not provide it either instead of getting it manually and setting it as default.
For example, we can receive a URL like `https:/01hnj9r1xrx3a4t3kb1ec7qs2b.api.us-east-1.staging.firebolt.io`, where account_id is included in the endpoint itself, so there is no need to add a query parameter as well

src/main/java/com/firebolt/jdbc/statement/rawstatement/StatementValidatorFactory.java
ptiurin, line 1: What's the idea behind this factory, why do we need it?
I understand what it's doing, but please explain why do we need it for this task?

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
stepansergeevitch: Ok, as for parameter makes sense, didn't see that.
As for account_id, my point is that we need to forcefully unset it (even if it was provided by a user). As I understand from the code, account_id value can still exist in the url even if it wasn't provided in system engine url response. Let me know if I'm missing someting

alexradzin: This logic is implemented in StatemtentClientImpl: lines 310-331

stepansergeevitch: Ok, makes sense

stepansergeevitch (approved):

PR #373, opened 2024-04-04
FIR-32147: now DatabaseMetadata of system connection behave exactly a…

alexradzin:
…s of user connection


alexradzin:
better implementation will be done instead

PR #374, opened 2024-04-04
Fir 32148 resultset getters case insensitive

alexradzin:


src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
```diff
@@ -175,14 +175,27 @@ public ResultSet getSchemas() throws SQLException {
 
 	@Override
 	public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLException {
-		String dbName = connection.getSessionProperties().getDatabase();
-		List<String> publicRow = Arrays.asList(PUBLIC_SCHEMA_NAME, dbName);
-		List<String> informationSchemaRow = Arrays.asList(INFORMATION_SCHEMA_NAME, dbName);
-		List<String> catalogRow = Arrays.asList(CATALOG_SCHEMA_NAME, dbName);
+		String catalogClause = catalog == null ? null : format("%s LIKE '%s'", TABLE_CATALOG, catalog);
+		String schemaClause = schemaPattern == null ? null : format("TABLE_SCHEMA LIKE '%s'", schemaPattern);
```
ptiurin, line 179: Should `TABLE_SCHEMA` also be a constant like `TABLE_CATALOG`?

src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
```diff
@@ -175,14 +175,27 @@ public ResultSet getSchemas() throws SQLException {
...
+		String where = Stream.of(catalogClause, schemaClause).filter(Objects::nonNull).collect(joining(" AND "));
+		if (where.isEmpty()) {
+			where = " WHERE " + where;
+		}
```
ptiurin, line 183: Should this if be an inverse instead? Otherwise if where is empty we're creating a broken SQL.

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -1242,6 +1243,29 @@ void shouldReturnUrl() throws SQLException, MalformedURLException {
...
+	@Test
+	void shouldBeCaseInsensitive() throws SQLException {
+		inputStream = getInputStreamWithCommonResponseExample();
+		resultSet = new FireboltResultSet(inputStream, "a_table", "a_db", 65535);
+		resultSet.next();
+		ResultSetMetaData rsmd = resultSet.getMetaData();
+		int n = rsmd.getColumnCount();
+		for (int i = 1; i <= n; i++) {
+			String columnName = rsmd.getColumnName(i);
+			System.out.println(columnName + ", " + rsmd.getColumnType(i));
```
ptiurin, line 1255: Let's not print in tests to avoid unnecessary output.

PR #375, opened 2024-04-08
FIR-32147: fixed DatabaseMetadata when connecting to system engine

alexradzin:


src/main/java/com/firebolt/jdbc/metadata/FireboltSystemEngineDatabaseMetadata.java
```diff
@@ -10,34 +9,38 @@
...
 	}
 
 	@Override
 	public boolean isReadOnly() throws SQLException {
 		return true;
 	}
+
+	private String fixSchema(String schemaPattern) {
+		return schemaPattern == null || INFORMATION_SCHEMA.contains(schemaPattern.replace("%", "").toLowerCase()) ? INFORMATION_SCHEMA : "does_not_exist";
+	}
```
ptiurin, line 45: One issue I see here is that pattern "a%b" will not work.
Why don't we keep % in here? I believe Firebolt supports % and _ in like : https://docs.firebolt.io/sql-reference/functions-reference/like.html

src/main/java/com/firebolt/jdbc/metadata/FireboltSystemEngineDatabaseMetadata.java
ptiurin, line 1: Can we also add some basic integration tests here?
e.g. run `getSchemas()` and make sure it contains public and information_schema; Create table and check that it's in `getTables()`
This way we'll have more confidence in this code than just mocks.

src/main/java/com/firebolt/jdbc/metadata/FireboltSystemEngineDatabaseMetadata.java
alexradzin: System engine can work with information_schema only. Attempt to access tables from public schema fail. This is the reason that I force to access information_schema only.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -517,7 +517,7 @@ public String nativeSQL(String sql) throws SQLException {
 	@Override
 	public boolean isReadOnly() throws SQLException {
 		validateConnectionIsNotClose();
-		return false;
+		return sessionProperties.isSystemEngine();
```
ptiurin, line 520: Can we claim read-only if we can create tables via system engine? Should we?

src/main/java/com/firebolt/jdbc/metadata/FireboltSystemEngineDatabaseMetadata.java
ptiurin: Ok I see. And to confirm, this is when system engine is connected with no database, right? Because otherwise when it's connected to a database we can query any tables.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
alexradzin: hmm. you are right. Let's return false.

src/main/java/com/firebolt/jdbc/metadata/FireboltSystemEngineDatabaseMetadata.java
ptiurin: As confirmed over IM we can't, I was confused. lgtm then.

ptiurin (approved):

ptiurin (approved):
apart from the comment above, lgtm

PR #381, opened 2024-04-11
FIR-31966 implemented nativeSQL()

alexradzin:


ptiurin (commented):
I don't think we should be doing this, as this is misleading. We're not actually converting sql so if someone passes incompatible sql statement from a different grammar we'll just return them the same thing claiming it's "native".
I think it's fine to throw an exception here, even SQL Server does not support it: https://learn.microsoft.com/en-us/sql/connect/jdbc/reference/nativesql-method-sqlserverconnection?view=sql-server-ver16

alexradzin:
I think there was some kind of misunderstanding here. 
According to [spec](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#:~:text=nativeSQL,-String%20nativeSQL(String&text=sql)%20throws%20SQLException-,Converts%20the%20given%20SQL%20statement%20into%20the%20system's%20native%20SQL,the%20driver%20would%20have%20sent.) `nativeSQL()` should never throw `SQLFeatureNotSupportedException`. 

This method is related to `Statement.setEscapeProcessing()` and should process [JDBC escape syntax](https://docs.oracle.com/javadb/10.8.3.0/ref/rrefjdbc1020262.html). I have opened the following separate [issue](https://packboard.atlassian.net/browse/FIR-32122) for this. 

However following your comment I've changed the implementation. Now exception is thrown only if currently unsupported escaping is needed.

ptiurin:
> I think there was some kind of misunderstanding here.

Yeah, I assumed this actually translates the sql dialect. It'll be great in the future to add an explanation to the PR description, as even when looking at JDBC spec I was not clear why we're doing this change.

ptiurin (approved):

PR #384, opened 2024-04-14
FIR-31965: setTransactionIsolation throws exception for unsupported level

alexradzin:


ptiurin (approved):
lgtm.
Let's change the PR title to something readable without line break before merging.

PR #385, opened 2024-04-14
FIR-31968 implemented Statement.setMaxFieldSize()

alexradzin:


src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
```diff
@@ -33,17 +33,19 @@ public class FireboltStatementService {
...
+	 * @param maxFieldSize         max field size
 	 * @param standardSql          indicates if standard sql should be used
 	 * @param systemEngine         indicates if system engine is used
 	 * @param statement           the statement
 	 * @return an InputStream with the result
 	 */
+	@SuppressWarnings("java:S107") //Number of parameters (8) > max (7). Not nice but probably not so bad for internal class
 	public Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper,
-									   FireboltProperties properties, int queryTimeout, int maxRows, boolean standardSql, boolean systemEngine, FireboltStatement statement)
+									   FireboltProperties properties, int queryTimeout, int maxRows, int maxFieldSize, boolean standardSql, boolean systemEngine, FireboltStatement statement)
```
ptiurin, line 44: I feel like suppressing this warning everywhere is not great, even in internal classes. Having methods with many parameters can impact maintainability and readability of the code (see PK for an example of number of parameters getting out of hand).
In here, is it possible to create a struct/class `ExecuteProperties` and store `queryTimeout`, `maxRows`, `maxFieldSize`, `standardSql`, and possibly `systemEngine` flags there? Or use some other method of keeping parameters manageable.

src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
```diff
@@ -106,15 +108,16 @@ private FireboltResultSet() {
 		columns = new ArrayList<>();
 		statement = null;
 		maxRows = 0;
+		maxFieldSize = 0;
```
ptiurin, line 111: nit: add a comment saying that 0 is unlimited.

ptiurin (commented):
Otherwise, looks good.

src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
alexradzin: You are absolutely right. I have done this in separate PR because this change required a lot of changes in tests.

ptiurin (approved):
Let's add that comment, then it's ok to merge.

src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
alexradzin: added comment

PR #389, opened 2024-04-15
FIR-32353: refactored FireboltStatementService.execute()

alexradzin:
number of parameters is reduced because some of them can be retrieved from passed statement

(*) most of change here are in test mockup. The real change is in `FireboltStatementService.execute()` and in place where this method is called.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
+++ b/src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
@@ -31,21 +31,18 @@ public class FireboltStatementService {
 	 *
 	 * @param statementInfoWrapper the statement info
 	 * @param properties the connection properties
-	 * @param queryTimeout         query timeout
-	 * @param maxRows              max rows
-	 * @param maxFieldSize         max field size
 	 * @param standardSql          indicates if standard sql should be used
-	 * @param systemEngine         indicates if system engine is used
 	 * @param statement           the statement
 	 * @return an InputStream with the result
 	 */
-	@SuppressWarnings("java:S107") //Number of parameters (8) > max (7). Not nice but probably not so bad for internal class
 	public Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper,
-									   FireboltProperties properties, int queryTimeout, int maxRows, int maxFieldSize, boolean standardSql, boolean systemEngine, FireboltStatement statement)
+									   FireboltProperties properties, boolean standardSql, FireboltStatement statement)
 			throws SQLException {
+		int queryTimeout = statement.getQueryTimeout();
+		boolean systemEngine = properties.isSystemEngine();
 		InputStream is = statementClient.executeSqlStatement(statementInfoWrapper, properties, systemEngine, queryTimeout, standardSql);
 		if (statementInfoWrapper.getType() == StatementType.QUERY) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
```diff
@@ -31,21 +31,19 @@ public class FireboltStatementService {
...
 	 * @param properties the connection properties
-	 * @param queryTimeout         query timeout
-	 * @param maxRows              max rows
-	 * @param maxFieldSize         max field size
 	 * @param standardSql          indicates if standard sql should be used
-	 * @param systemEngine         indicates if system engine is used
 	 * @param statement           the statement
 	 * @return an InputStream with the result
 	 */
 	@SuppressWarnings("java:S107") //Number of parameters (8) > max (7). Not nice but probably not so bad for internal class
```
ptiurin, line 38: Let's un-suppress this warning now.

src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
```diff
@@ -63,11 +61,11 @@ public boolean isStatementRunning(String statementLabel) {
...
 	}
 
-	private FireboltResultSet createResultSet(InputStream inputStream, QueryRawStatement initialQuery, FireboltProperties properties, FireboltStatement statement, int maxRows, int maxFieldSize)
+	private FireboltResultSet createResultSet(InputStream inputStream, QueryRawStatement initialQuery, FireboltProperties properties, FireboltStatement statement)
 			throws SQLException {
 		return new FireboltResultSet(inputStream, Optional.ofNullable(initialQuery.getTable()).orElse(UNKNOWN_TABLE_NAME),
 				Optional.ofNullable(initialQuery.getDatabase()).orElse(properties.getDatabase()),
-				properties.getBufferSize(), maxRows, maxFieldSize, properties.isCompress(), statement,
+				properties.getBufferSize(), statement.getMaxRows(), statement.getMaxFieldSize(), properties.isCompress(), statement,
 				properties.isLogResultSet());
```
ptiurin, line 69: Can this solution be extended to `FireboltResultSet`? We're doing the same thing, passing statement but also max rows and field size.

alexradzin: done

src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
alexradzin: removed

src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
ptiurin: Does this mean we can remove the old constructor? https://github.com/firebolt-db/jdbc/pull/389/files#diff-62a56c126996c1cf26d2c0ef1e090463acd6ffc55d127af126857ea350507147R124-R126

alexradzin: I've reduced access modifiers of some constructors. Unfortunately removing some constructors requires massive changes in tests. I will open separate issue for this. Otherwise the changes will look irrelevant for this PR.

ptiurin (approved):

alexradzin:
retest

PR #391, opened 2024-04-16
FIR-32019: PreparedStatement.setObject with all parameters

alexradzin:
also added more tests of prepared statement

### DIFF
--- a/src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
+++ b/src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
@@ -8,13 +8,17 @@ import com.firebolt.jdbc.type.date.SqlDateUtil;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.Array;
+import java.sql.Blob;
+import java.sql.Clob;
 import java.sql.Date;
+import java.sql.JDBCType;
+import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.UUID;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
```diff
@@ -24,23 +28,53 @@
...
 	STRING(String.class, getSQLStringValueOfString()),
-	LONG(Long.class, String::valueOf),
-	INTEGER(Integer.class, String::valueOf),
-	BIG_INTEGER(BigInteger.class, String::valueOf),
-	FLOAT(Float.class, String::valueOf),
-	DOUBLE(Double.class, String::valueOf),
+	LONG(Long.class, value -> Long.toString(((Number)value).longValue())),
+	INTEGER(Integer.class, value -> Integer.toString(((Number)value).intValue())),
+	BIG_INTEGER(BigInteger.class, value -> value instanceof BigInteger ? value.toString() : Long.toString(((Number)value).longValue())),
+	FLOAT(Float.class, value -> Float.toString(((Number)value).floatValue())),
```
ptiurin, line 37: Can you explain why did we switch from `valueOf`? What's the advantage here?

ptiurin (approved):
Otherwise, looks ok to my knowledge. Did integration tests succeed?

src/main/java/com/firebolt/jdbc/type/JavaTypeToFireboltSQLString.java
alexradzin: ValueOf just calls `toString()`. I wanted to be sure that the passed value (of type Object) indeed can be transformed to the required type. This never happens if you use typed setters like setInt() or setDouble() but can happen when you use setObject(). For example one can say

```java
ps.setObject(1, "one", Types.INTEGER); 
```

The compilation will obviously work but we want to fail at runtime in this line and not wait for the statement execution.

PR #394, opened 2024-04-17
FIR-32016, FIR-32018: implemented blob, clob, stream in PS

alexradzin:
Blob, clob and stream related methods in prepared statement


src/main/java/com/firebolt/jdbc/util/InputStreamUtil.java
```diff
@@ -27,4 +28,17 @@ public void readAllBytes(@Nullable InputStream is) {
             }
         }
     }
+
+    public String read(Reader initialReader, int limit) throws IOException {
+        char[] arr = new char[8 * 1024];
```
ptiurin, line 33: Let's have this as a constant.

src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
ptiurin, line 1: Let's add integration tests as well to cover these types.

ptiurin (approved):

PR #396, opened 2024-04-18
FIR-32017: PreparedStatement: set date/time/timestamp with calendar

alexradzin:
also added simple setTime() as a pair to the ResultSet.getTime()

### DIFF
--- a/src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
+++ b/src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
@@ -319,9 +319,8 @@ public class FireboltPreparedStatement extends FireboltStatement implements Prep
 	}
 
 	@Override
-	@NotImplemented
-	public void setDate(int parameterIndex, Date x, Calendar cal) throws SQLException {
-		throw new FireboltUnsupportedOperationException();
+	public void setDate(int parameterIndex, Date date, Calendar calendar) throws SQLException {
+		setDateTime(parameterIndex, date, calendar, JavaTypeToFireboltSQLString.DATE);
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
```diff
@@ -171,9 +171,10 @@ public void setDate(int parameterIndex, Date x) throws SQLException {
 	}
 
 	@Override
-	@NotImplemented
 	public void setTime(int parameterIndex, Time x) throws SQLException {
-		throw new SQLFeatureNotSupportedException("The format Time is currently not supported");
+		validateStatementIsNotClosed();
+		validateParamIndex(parameterIndex);
+		providedParameters.put(parameterIndex, JavaTypeToFireboltSQLString.TIME.transform(x));
```
ptiurin, line 177: I'm not sure about this. Firebolt does not support Time type so we shouldn't mislead users into allowing to set it.

alexradzin: This is exactly as int, nstring or blob. We do not support blobs as well. Right now I treat blob as a byte array and allow to use Time on both prepared statement and result set. In fact Time extends Date and in implementations that I know stores epoch in milliseconds. The translation from epoch to Time (that knows hour, minutes, seconds, milliseconds but doesn't "know" date is done as a kind of simulation. 

Moreover according to the spec this method should not throw SQLFeatureNotSupportedException, so the driver should somehow provide implementation. This is the best implemnentation I can do now, but probably in future Firebolt will support this type internally, so we will be able to change the implementation.

ptiurin: As discussed on the daily, I see the point here, but let's have the product team make a decision what's best for the users.

alexradzin: setTime() is removed, please review again

ptiurin (commented):
Otherwise looks good

ptiurin (approved):

PR #399, opened 2024-04-22
FIR-32538: optimisation of FireboltResultSet constructors

alexradzin:


src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
```diff
@@ -154,25 +115,20 @@ private FireboltResultSet(InputStream is, String tableName, String dbName, Integ
...
 		InputStreamReader inputStreamReader;
 		if (isCompressed) {
 			inputStreamReader = new InputStreamReader(new LZ4InputStream(is), UTF_8);
 		} else {
 			inputStreamReader = new InputStreamReader(is, UTF_8);
 		}
-		return bufferSize != null && bufferSize != 0 ? new BufferedReader(inputStreamReader, bufferSize)
-				: new BufferedReader(inputStreamReader);
+		return new BufferedReader(inputStreamReader, bufferSize);
 	}
```
ptiurin, line 132: Reading the BufferedReader documentation, it throws an exception when buffer size is 0. We used to account for this by setting it with no buffer size (using the default) but now we'll get an exception in this case. 
Should we be throwing it?

ptiurin (commented):
Apart from this looks good.

src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
ptiurin: I'll need a bit more explanation. Does setting it in FireboltProperties guarantee we'll never receive 0 here?

alexradzin: IMHO yes. Moreover we will never get null there. Take a look on the constructor `BufferedReader(Reader)`. It calls `BufferedReader(Reader, defaultBufferSize)` while `defaultBufferSize=8K`. So,  the only thing I did is the code simplification. We always provide reasonable `bufferSize`.

ptiurin: Discussed offline, should never happen. In any case buffer size 0 should fail.

ptiurin (approved):

PR #401, opened 2024-05-01
fixed nullable setURL() test and nullable fields in integration tests

alexradzin:


src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
```diff
@@ -147,7 +147,7 @@ void getTables(String catalog, String schemaPattern, String tableNamePattern, St
             ",%schema,%in%,table_name,false",
             ",%form%,%in%,table_schema,false",
             "{database},,%in%,table_catalog,false",
-            "{database},information_schema,%in%,account_id,false",
+            "{database},information_schema,%ac%,account_id,false",
             "{database},information%,%in%,name,false",
```
ptiurin, line 151: I'm not sure I understand, what does this do?

ptiurin (approved):
Just one q, otherwise lgtm.

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
alexradzin: This exactly what I mentioned. Once upon a time this test found table named `running_queries` that is now not accessible from system engine. However pattern `%ac%` matches 2 tables: `accounts` and `rbac_roles`. I will modify this test and add expected tables instead of `emptyList=true|false`

PR #407, opened 2024-05-12
removed redundant flag standardSql and added more tests

alexradzin:


src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -101,10 +100,10 @@ protected Optional<ResultSet> execute(List<StatementInfoWrapper> statements) thr
 	}
 
 	@SuppressWarnings("java:S2139") // TODO: Exceptions should be either logged or rethrown but not both
-	private Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper, boolean verifyNotCancelled, boolean isStandardSql) throws SQLException {
+	private Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper) throws SQLException {
 		createValidator(statementInfoWrapper.getInitialStatement(), connection).validate(statementInfoWrapper.getInitialStatement());
 		ResultSet resultSet = null;
-		if (!verifyNotCancelled || isStatementNotCancelled(statementInfoWrapper)) {
```
ptiurin, line 107: What does `verifyNotCancelled` flag do here? Are we safe to remove it?

PR #410, opened 2024-05-13
FIR-32995 fixed numeric types transformations

alexradzin:


src/main/java/com/firebolt/jdbc/type/BaseType.java
```diff
@@ -19,16 +18,20 @@
...
 public enum BaseType {
-	LONG(Long.class, conversion -> Long.parseLong(checkInfinity(conversion.getValue()))),
-	INTEGER(Integer.class, conversion -> Integer.parseInt(checkInfinity(conversion.getValue()))),
-	SHORT(Short.class, conversion -> Short.parseShort(checkInfinity(conversion.getValue()))),
-	BIGINT(BigInteger.class, conversion -> new BigInteger(checkInfinity(conversion.getValue()))),
+	LONG(TypePredicate.mayBeFloatingNumber, Long.class, conversion -> Long.parseLong(checkInfinity(conversion.getValue())), conversion -> Double.valueOf(conversion.getValue()).longValue()),
+	INTEGER(TypePredicate.mayBeFloatingNumber, Integer.class, conversion -> Integer.parseInt(checkInfinity(conversion.getValue())), conversion -> Double.valueOf(conversion.getValue()).intValue()),
+	SHORT(TypePredicate.mayBeFloatingNumber, Short.class, conversion -> Short.parseShort(checkInfinity(conversion.getValue())), conversion -> Double.valueOf(conversion.getValue()).shortValue()),
+	BYTE(TypePredicate.mayBeFloatingNumber, Byte.class, conversion -> Byte.parseByte(checkInfinity(conversion.getValue())), conversion -> Double.valueOf(conversion.getValue()).byteValue()),
```
ptiurin, line 33: Why do we have `Double.valueOf` in the second function in each type instead of the actual type?

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -1058,15 +1060,47 @@ void shouldGetObjectsForNumericTypes() throws SQLException {
...
+		assertEquals((byte)1, resultSet.getObject(4, Byte.class));
+		assertEquals((short)1, resultSet.getObject(4, Short.class));
+		assertEquals(1, resultSet.getObject(4, Integer.class));
+		assertEquals(1, resultSet.getObject(4, Long.class));
+		assertEquals(1.23456789012f, resultSet.getObject(4, Float.class));
 		assertEquals(1.23456789012, resultSet.getObject(4, Double.class));
 		assertEquals(new BigDecimal("1.23456789012"), resultSet.getObject(4, BigDecimal.class));
+
+		//assertEquals((byte)1, resultSet.getObject(5, Byte.class));
+		//assertEquals((short)1231232, resultSet.getObject(5, Short.class));
```
ptiurin, line 1099: Why do we have these commented out? Should they be working?

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -1058,15 +1060,47 @@ void shouldGetObjectsForNumericTypes() throws SQLException {
...
+		assertEquals(1L, resultSet.getObject(1, Long.class));
 		assertEquals(new BigInteger("1"), resultSet.getObject(1, BigInteger.class));
 		assertEquals(1, resultSet.getObject(1, Long.class));
+		assertEquals(1, resultSet.getObject(1, Float.class));
+		assertEquals(1, resultSet.getObject(1, Double.class));
+
+		// the number is too big
+		FireboltException e = assertThrows(FireboltException.class, () -> resultSet.getObject(2, Integer.class));
+		assertEquals(TYPE_TRANSFORMATION_ERROR, e.getType());
+		assertEquals(NumberFormatException.class, e.getCause().getClass());
```
ptiurin, line 1075: Let's get more of these negative tests, e.g. int -> short, int -> byte to make sure we're always failing for all these cases.

src/main/java/com/firebolt/jdbc/type/BaseType.java
```diff
@@ -159,8 +174,22 @@ private <T> T transform(StringToColumnTypeConversion conversion) throws SQLExcep
...
+					throw new FireboltException(e.getMessage(), e, TYPE_TRANSFORMATION_ERROR);
+				}
+			}
+		}
+		try {
+			//noinspection unchecked
+			return (T) transformFunctions[transformFunctions.length - 1].apply(conversion);
+		} catch (RuntimeException e) {
+			throw new FireboltException(e.getMessage(), e, TYPE_TRANSFORMATION_ERROR);
+		}
```
ptiurin, line 192: Can't we merge the for loop and the last transform case? It should be just the matter of modifying the if condition. I think this way it will be easier to understand.

src/main/java/com/firebolt/jdbc/type/BaseType.java
```diff
@@ -124,9 +139,9 @@ private static boolean isNan(String value) {
 		return "nan".equalsIgnoreCase(value) || "+nan".equalsIgnoreCase(value) || "-nan".equalsIgnoreCase(value);
 	}
 
-	private static void validateObjectNotNull(String value) {
+	private static void validateObjectNotNull(String value) throws SQLException {
 		if (value == null) {
-			throw new IllegalArgumentException("The value cannot be null");
+			throw new FireboltException("The value cannot be null");
 		}
 	}
```
ptiurin, line 146: Let's try improving the error message here since by using FireboltException we're removing some of the information. Perhaps adding what operation went wrong would be better?

alexradzin: reverted. In fact this cannot happen because value cannot be null here.

alexradzin: Although I disagree that replacing IllegalArgumentException by FireboltException (i.e. SQLException) looses any information.

src/main/java/com/firebolt/jdbc/type/BaseType.java
alexradzin: Done. You are right - this version is simpler

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
alexradzin: fixed test and the code. 10x

src/main/java/com/firebolt/jdbc/type/BaseType.java
alexradzin: The second function can convert floating point number to specific integer type (byte, short, int, long). So it has to parse the number as floating point (either Float or Double) and then take it integer part only. Double is preferable here because it can work with bigger number and it is the java default. Type Float was introduced into java 30 years ago when computers had 16M RAM and 1G disk.

ptiurin: Ok, makes sense.

src/main/java/com/firebolt/jdbc/type/BaseType.java
ptiurin: I think the information that the argument caused this error is helpful (at least partially). FireboltException class is generic and can fit many errors, including some backend ones, while IllegalArgumentException points at the argument provided. I'm not saying one is preferable over another, but that it helps narrowing down where the issue is coming from.
In any case, it's a minor point.

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -1053,21 +1056,124 @@ void shouldReturnDataForNewNonNumericDataTypes() throws SQLException {
 		assertNull(resultSet.getObject(13));
 	}
 
+	@SuppressWarnings("java:S5961") // all assertions are necessary
```
ptiurin, line 1059: They are necessary, but the SC issue here is with having a huge single "god" test. I'd rather not suppress thing needlessly and refactor this into two (or more) test cases. One I can think of is having all errors test, e.g. `shouldErrorOnTooLargeNumericTypes`. Or even separating it into `shouldGetObjectsForByte`, `shouldGetObjectsForShort` , etc. if the previous suggestion is not enough.

ptiurin (commented):
Otherwise looks good.

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
alexradzin: Done. Now we have separate functions for each type.

ptiurin (approved):

PR #417, opened 2024-05-28
FIR-33264: quoted engine and DB names

alexradzin:


stepansergeevitch (commented):
Please add a test for v2 which
- Creates an engine with uppercase letters in names
- Tries to connect to it
Otherwise LGTM

stepansergeevitch (approved):

PR #424, opened 2024-06-10
feat: FIR-33601: use catalogs table instead of databases table if exists

alexradzin:


src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -17,29 +16,38 @@
...
     private static final Collection<String> RUNNING_STATUSES = Stream.of("running", "ENGINE_STATE_RUNNING").collect(toCollection(() -> new TreeSet<>(CASE_INSENSITIVE_ORDER)));
     private static final String ENGINE_QUERY =
-            "SELECT engs.url, engs.attached_to, dbs.database_name, engs.status, engs.engine_name " +
+            "SELECT engs.url, engs.attached_to, dbs.%1$s_name, engs.status, engs.engine_name " +
             "FROM information_schema.engines as engs " +
-            "LEFT JOIN information_schema.databases as dbs ON engs.attached_to = dbs.database_name " +
+            "LEFT JOIN information_schema.%1$ss as dbs ON engs.attached_to = dbs.%1$s_name " +
             "WHERE engs.engine_name = ?";
-    private static final String DATABASE_QUERY = "SELECT database_name FROM information_schema.databases WHERE database_name=?";
+    private static final String DATABASE_QUERY = "SELECT %1$s_name FROM information_schema.%1$ss WHERE %1$s_name=?";
```
stepansergeevitch, line 29: Should this be called IS_QUERY or something? Since it's not only for querying databases

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
```diff
@@ -17,29 +16,38 @@
...
 
     private final FireboltConnection fireboltConnection;
+    private final String dbTerm;
+
+    public FireboltEngineInformationSchemaService(FireboltConnection fireboltConnection) throws SQLException {
+        this.fireboltConnection = fireboltConnection;
+        dbTerm = doesRecordExist(format(DATABASE_QUERY, "table"), "catalogs") ? "catalog" : "table";
```
stepansergeevitch, line 36: Should we use `databases` in case we don't have catalogs? I don't think `table` makes sense here

alexradzin: I do not know. I tried to be backwards compatible. Do you think that we should just forget about the `databases` table for all environments? BTW if this is correct the all already released drivers are broken.

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
alexradzin: I agree that now the name is not good enough, however IMHO `IS_QUERY` even worse. What about `EXISTENCE_QUERY` or, probably `INVENTORY_QUERY`?
But taking in consideration your next comment all this is irrelevant and we just have to  use catalogs instead of databases.

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
stepansergeevitch: But we are here checking if `information_schema.catalogs` table exists. If it does, we continue to use it. If not, we choose to use `information_schema.table`, which I believe is invalid and it should be `information_schema.databases`. Am I missing something?

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
stepansergeevitch: We can use both suggested names. I'm just thinking to make the naming more general

alexradzin: Renamed to `INVENTORY_QUERY`.

src/main/java/com/firebolt/jdbc/service/FireboltEngineInformationSchemaService.java
alexradzin: You are right; I've fixed a bug. Please review again.

stepansergeevitch (approved):

PR #426, opened 2024-06-16
FIR-33627: json formatted server error

alexradzin:


src/main/java/com/firebolt/jdbc/client/FireboltClient.java
```diff
@@ -194,6 +203,49 @@ private String extractErrorMessage(Response response, boolean isCompress) throws
...
+			}
+			return new ServerError(null, new ServerError.Error[] {new ServerError.Error(null, message, null, null, null, null, null, location)});
+		}
+	}
+
+	private ServerError.Error updateError(ServerError.Error error) {
+		if (error == null || error.getDescription() == null) {
+			return error;
+		}
+		Entry<Location, String> locationAndText = getLocationFromMessage(error.getDescription());
```
stepansergeevitch, line 230: I see that we call `updateError` in case where the error is in a JSON format. Yet I see that we still take the location from the description, while I believe we should take one from the actual `location` json field. Am I missing something?

src/main/java/com/firebolt/jdbc/exception/ServerError.java
```diff
@@ -0,0 +1,295 @@
...
+            return description;
+        }
+
+        public String getResolution() {
+            return resolution;
+        }
+
+        public String getHelpLink() {
+            return helpLink;
+        }
```
ptiurin, line 183: Can we use something like lombok's [getters](https://projectlombok.org/features/GetterSetter) to avoid writing out every single parameter here?
Also possibly `@EqualsAndHashCode`

alexradzin: We can but I  planned to remove Lombok, so it looks strange to continue using it.

src/main/java/com/firebolt/jdbc/client/FireboltClient.java
alexradzin: Yes, the case is that right now the location does not present in returned JSON but it can be found in the description. This was the reason to implement this logic. If location is NOT found into description the error is returned as-is (i.e. including location). Otherwise location extracted from the description is used.

alexradzin: Here is the example of JSON that I got from the server

```
{
  "errors": [
    {
      "description": "Line 1, Column 8: Column 'wrong' does not exist."
    }
  ],
  "query": {
    "query_id": "77f45647-305e-4611-9537-81bc081dc3f6",
    "request_id": "c263509e-b6b2-49db-8c9b-8dcf274b68c9"
  },
  "statistics": {
    "elapsed": 0.0
  }
}
```

src/main/java/com/firebolt/jdbc/exception/ServerError.java
alexradzin: You are right, it does not affect the package wight but IMHO sometimes it makes code worse. I can explain more but if you really want I can add the annotation.

src/main/java/com/firebolt/jdbc/client/FireboltClient.java
stepansergeevitch: According to the design doc, I see that location can be returned as a part of json in `location` field. Do we check if that is present?
```
{
      "code": "CMP0001",
      "name": "<error_name>",
      "severity": "ERROR",
      "source": "User Error",
      "description": "Division by zero",
      "resolution": "",
      "helpLink": "",
      "location": {
        "failingLine": 42,
        "startOffset": 120,
        "endOffset": 135
      }
    }
    ```

src/main/java/com/firebolt/jdbc/exception/ServerError.java
ptiurin: Right, so let's use it appropriately in places where it doesn't make the code worse. Here for example it will be beneficial IMO.
Let's do big design decisions (like removing some dependencies) properly, with a doc or ticket describing the reasoning that's also shared and discussed with the team.

src/main/java/com/firebolt/jdbc/client/FireboltClient.java
alexradzin: Exactly. This does not present so far in my tests. I posted here the real JSON that I got from the server. IMHO my implementation supports both cases.

src/main/java/com/firebolt/jdbc/exception/ServerError.java
alexradzin: After a long discussion I have added Lombok annotations there

src/main/java/com/firebolt/jdbc/client/FireboltClient.java
alexradzin: Anyway, I have fixed thew logic: if Location exists its value will be used. Only if Location is null but column and line can be extracted from the message we use this information.

stepansergeevitch (approved):

PR #432, opened 2024-06-28
feat(FIR-34158): populating SQLState in firebolt errors

ptiurin:
FireboltException is probably a new prime candidate for a Builder, as the number of constructors is getting out of hand.

SQLStates are copied from [wiki](https://en.wikipedia.org/wiki/SQLSTATE). Focus here on implementing the ones [needed](https://tableau.github.io/connector-plugin-sdk/docs/drivers.html#important-sqlstates) for Tableau.


src/main/java/com/firebolt/jdbc/service/FireboltAuthenticationService.java
```diff
@@ -51,7 +53,11 @@ public FireboltConnectionTokens getConnectionTokens(String host, FireboltPropert
 		} catch (FireboltException e) {
 			log.log(Level.SEVERE, "Failed to connect to Firebolt", e);
 			String msg = ofNullable(e.getErrorMessageFromServer()).map(m -> format(ERROR_MESSAGE_FROM_SERVER, m)).orElse(format(ERROR_MESSAGE, e.getMessage()));
-			throw new FireboltException(msg, e);
+			SQLState sqlState = SQLState.fromCode(e.getSQLState());
+			if (sqlState == null) {
```
stepansergeevitch, line 57: nit: Do we need this if? Can we just put whatever value we get as a code?

stepansergeevitch (approved):

src/main/java/com/firebolt/jdbc/exception/SQLState.java
```diff
@@ -0,0 +1,125 @@
+package com.firebolt.jdbc.exception;
+
+public enum SQLState {
```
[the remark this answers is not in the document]
ptiurin: Initially had a link to a competing db vendor so I didn't put it here, but turns out there's also a wikipedia article. Added a link to it in the code.

src/main/java/com/firebolt/jdbc/service/FireboltAuthenticationService.java
ptiurin: You're right. I had an issue with null value before so I had to add this check. However, with a fresh mind today I've found an arguably neater solution.

stepansergeevitch (approved):

PR #448, opened 2024-07-30
fix: Fir 35049 dont reset engine autostop with service queries in drivers for jdbc

stepansergeevitch:
Added auto_start_stop_control=ignore parameter to isValid call of JDBC connection


stepansergeevitch:
https://github.com/firebolt-db/jdbc/actions/runs/10160971649

ptiurin (approved):

PR #452, opened 2024-09-02
refactor: Fir 35469 remove account resolve endpoint from jdbc

stepansergeevitch:
Removed account resolve call. Still leaving infraVersion variable in place since it's currently used to distinguish between fb1.0 and 2.0. Refactoring it out would be a much bigger effort

### DIFF
--- a/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
+++ b/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
@@ -314,9 +314,12 @@ public class StatementClientImpl extends FireboltClient implements StatementClie
 			}
 		} else {
 			if (connection.getInfraVersion() >= 2) {
+				String engine = fireboltProperties.getEngine();
 				if (accountId != null) {
 					params.put(FireboltQueryParameterKey.ACCOUNT_ID.getKey(), accountId);
-					params.put(FireboltQueryParameterKey.ENGINE.getKey(), fireboltProperties.getEngine());
+				}
+				if (engine != null) {
+					params.put(FireboltQueryParameterKey.ENGINE.getKey(), engine);
 				}
 				params.put(FireboltQueryParameterKey.QUERY_LABEL.getKey(), statementInfoWrapper.getLabel()); //QUERY_LABEL
 			}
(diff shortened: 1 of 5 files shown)


src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -314,9 +314,12 @@ private Map<String, String> getAllParameters(FireboltProperties fireboltProperti
...
 		} else {
 			if (connection.getInfraVersion() >= 2) {
+				String engine = fireboltProperties.getEngine();
 				if (accountId != null) {
 					params.put(FireboltQueryParameterKey.ACCOUNT_ID.getKey(), accountId);
-					params.put(FireboltQueryParameterKey.ENGINE.getKey(), fireboltProperties.getEngine());
+				}
+				if (engine != null) {
+					params.put(FireboltQueryParameterKey.ENGINE.getKey(), engine);
 				}
```
ptiurin, line 323: Why are we changing this?

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
```diff
@@ -102,19 +99,17 @@ protected void assertDatabaseExisting(String database) throws SQLException {
...
-        String accountId = systemEngineUrlUrlParams.getOrDefault(ACCOUNT_ID.getKey(), account.getId());
         for (Entry<String, String> e : systemEngineUrlUrlParams.entrySet()) {
             loginProperties.addProperty(e);
         }
         return loginProperties
                 .toBuilder()
                 .systemEngine(true)
                 .compress(false)
-                .accountId(accountId)
+                .accountId(null)
```
ptiurin, line 112: Can we not remove accountid line here altogether since this is a builder? Or are we wiping it on purpose?

stepansergeevitch: Just left it like this for visibility. Also, I'm not sure what default value is set, so setting this explicitly made sense to me

ptiurin: The default here is [null](https://github.com/firebolt-db/jdbc/blob/9daa67fc6274e339e6941438202af12412374044/src/main/java/com/firebolt/jdbc/connection/settings/FireboltSessionProperty.java#L63) so let's remove this line to avoid unnecessary complexity.

ptiurin (approved):
Apart from that looks good

PR #454, opened 2024-09-19
fix(FIR-35270): Avoid printing secrets in logs

ptiurin:
Sanitizing logs to avoid printing secrets. Bumping down query printing to debug level instead.
Also fixed skipped tests.


ptiurin:
Integration tests are [passing](https://github.com/firebolt-db/jdbc/actions/runs/10944524101), failures are consistent with master.

PR #464, opened 2024-10-17
fix(FIR-37280): parameter formatting for the new Firebolt

ptiurin:
New Firebolt no longer requires slashes to be escaped. Adding a change to detect version we're running on and use the appropriate escaping.


src/main/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatement.java
```diff
@@ -198,6 +206,14 @@ public void setObject(int parameterIndex, Object x, int targetSqlType) throws SQ
...
 		try {
+			if (this.getConnection().getClass() == FireboltConnectionUserPassword.class) {
+				// We don't know the targetSqlType, so we let JavaTypeToFireboltSQLString deal
+				// with legacy conversion
+				providedParameters.put(parameterIndex,
+						JavaTypeToFireboltSQLString.transformAny(x, targetSqlType, ParserVersion.LEGACY));
+			} else {
+				providedParameters.put(parameterIndex, JavaTypeToFireboltSQLString.transformAny(x, targetSqlType));
+			}
 			providedParameters.put(parameterIndex, JavaTypeToFireboltSQLString.transformAny(x, targetSqlType));
```
stepansergeevitch, line 217: Can we have a connection class have a parser_version variable? So that we can jsut pass it in the end.
Also, it seems like we do `put` twice, once in if and second time on line 217, is that intended?

ptiurin: > it seems like we do put twice
Well spotted, looks like I merged this with errors.

ptiurin: As discussed offline, refactored connection class detection.

stepansergeevitch (approved):

PR #473, opened 2024-10-29
Fir 37710 jdbc validate connection against system engine

stepansergeevitch:
Improvements for isValid method:
- FIR-37621 run validation for a system engine as well
- add a connection property `validate_on_system_engine`, which forces a connection to run isValid on a system engine even if it's connected to a user engine

### DIFF
--- a/src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
+++ b/src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
@@ -51,6 +51,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.IdentityHashMap;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -61,7 +62,6 @@ import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
-import static com.firebolt.jdbc.connection.FireboltConnectionUserPassword.SYSTEM_ENGINE_NAME;
 import static com.firebolt.jdbc.connection.settings.FireboltSessionProperty.getNonDeprecatedProperties;
 import static java.lang.String.format;
 import static java.sql.ResultSet.CLOSE_CURSORS_AT_COMMIT;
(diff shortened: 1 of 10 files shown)


src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -438,11 +437,15 @@ public boolean isValid(int timeout) throws SQLException {
...
 			throws SQLException {
-		FireboltProperties propertiesCopy = FireboltProperties.copy(fireboltProperties);
+		HashMap<String, String> runtimeProperties = new HashMap<>(fireboltProperties.getRuntimeAdditionalProperties());
 		if (isInternalRequest) {
-			propertiesCopy.addProperty("auto_start_stop_control", "ignore");
+			runtimeProperties.put("auto_start_stop_control", "ignore");
+		}
+		var propertiesBuilder = fireboltProperties.toBuilder().runtimeAdditionalProperties(runtimeProperties);
+		if (getSessionProperties().isValidateOnSystemEngine()) {
+			propertiesBuilder.compress(false).engine(null).systemEngine(true);
```
ptiurin, line 446: Why do we have `compress(false)` defined here? Is there a special behaviour on system engine?

src/main/java/com/firebolt/jdbc/connection/settings/FireboltSessionProperty.java
```diff
@@ -67,6 +67,8 @@ public enum FireboltSessionProperty {
 	USER_CLIENTS("user_clients", null, String.class, "user clients", FireboltProperties::getUserClients),
 	ACCESS_TOKEN("access_token", null, String.class, "access token", p -> "***"),
 	ENVIRONMENT("environment", "app", String.class, "Firebolt environment", FireboltProperties::getEnvironment, "env"),
+	VALIDATE_ON_SYSTEM_ENGINE("validate_on_system_engine", false, Boolean.class,
+			"Whether to validate the connection on the system engine or not. Not validated by default", FireboltProperties::isValidateOnSystemEngine),
```
ptiurin, line 71: I'd reword this to avoid confusion, otherwise it sounds like we're not doing validation at all. `"Whether to validate the connection on the system engine or not. By default validates on an engine currently connected. "`  or something similar.

ptiurin (approved):
Couple of small points, otherwise looks good.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
stepansergeevitch: For system engine we automatically disable compress in JDBC.
https://github.com/firebolt-db/jdbc/blob/master/src/main/java/com/firebolt/jdbc/connection/settings/FireboltProperties.java#L97
I'm not sure of the reasons, all that I can say is that system engine doesn't work with compres as tests have shown. We receive an expected end of stream, and I failed ot understand the reason behind it

PR #483, opened 2024-12-06
feat(FIR-38126): struct type

ptiurin:


src/main/java/com/firebolt/jdbc/resultset/column/ColumnType.java
```diff
@@ -108,6 +111,8 @@ private static List<ColumnType> getCollectionSubType(FireboltDataType fireboltDa
 		if (fireboltDataType.equals(TUPLE)) {
 			types = typeWithoutNullKeyword.split(",(?![^()]*\\))"); // Regex to split on comma and ignoring comma that are between
 			// parenthesis
+		} else if (fireboltDataType.equals(STRUCT)) {
+			types = typeWithoutNullKeyword.split(",");
```
stepansergeevitch, line 115: I believe this will fail for nested structs, for example
```
struct(a int, s struct(a int, b text))
typeWithoutNullKeyword.split(",") = "a int, s struct(a int, b text)".split(",") = ["a int", "s struct(a int", "b text)"]
```

src/test/resources/responses/firebolt-response-with-struct-nofalse
```diff
@@ -0,0 +1,4 @@
+id	null_struct	an_empty_struct	a_struct
+Int64	struct(a int null)	struct(a int)	struct(a int)
```
stepansergeevitch, line 2: Please add a nested struct and a struct with array here

src/main/java/com/firebolt/jdbc/resultset/column/ColumnType.java
ptiurin: Well spotted. Looks like it's already taken care of in regex above so I've stolen it.

stepansergeevitch (approved):

PR #491, opened 2024-12-23
fix: bytea in array parsing

stepansergeevitch:
Bytea, when returned in array in CSV, has a `\\x` prefix instead of a standard `\x`.

In the FireboltResultSetTest you will see that I have changed the method from getBytes to getString since the column name is of type String.

The way I read the JDBC spec (https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf?AuthParam=1744031434_3e1d01e62d033e97eb5e7c88f5039beb) table B6, we cannot use getBytes on the result set unless the column that we want to get they bytes for is of type (BINARY, VARBINARY, LONGVARBINARY). So changed the tests to getString instead of getBytes and getChararcterStream instead of getBinaryStream.


goprean (approved):

src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
```diff
@@ -5,24 +5,23 @@
...
+import lombok.CustomLog;
+import lombok.NonNull;
 
 import static java.lang.String.format;
-import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Optional.ofNullable;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 
+@SuppressWarnings("javaarchitecture:S7091") // was not added as part of this change and was already there
```
ptiurin, line 24: Let's resolve this one in the SonarCloud UI so for this PR it doesn't bother us, but is still shown as an issue in main. This way it's not sweeped under a rug and we can come back to it at some point.

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -519,18 +518,26 @@ void shouldReturnBytes() throws SQLException, IOException {
...
+		assertEquals(expectedString, readAll(resultSet.getCharacterStream(3)));
+		assertEquals(expectedString, readAll(resultSet.getCharacterStream("name")));
+		assertEquals(expectedString, resultSet.getObject(3, String.class));
+		assertEquals(expectedString, resultSet.getObject("name", String.class));
 		resultSet.next();
-		assertNull(resultSet.getBytes(3));
-		assertNull(resultSet.getBlob(3));
+		assertNull(resultSet.getString(3));
+		assertNull(resultSet.getString(3));
+	}
```
ptiurin, line 531: This test is `shouldReturnBytes` but we're not testing bytes anymore. If this is already tested elsewhere we can remove it?

src/main/java/com/firebolt/jdbc/type/array/SqlArrayUtil.java
goprean: Done.

ptiurin (approved):

PR #497, opened 2025-03-05
feat(FIR-34063):implement get catalogs to return all dbs

bogdantruta-firebolt:
Implemented the getCatalogs method from FireboltDatabaseMetadata to return all catalogs instead of just the one from the connection. Changed the existing unit test for this method and also created an integration test that checks that at least the current db provided for the integration tests is returned in the ResultSet (due to there being lots of databases available in staging)

### DIFF
--- a/src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
+++ b/src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
@@ -9,11 +9,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 
-import java.sql.Connection;
-import java.sql.DatabaseMetaData;
-import java.sql.ResultSet;
-import java.sql.ResultSetMetaData;
-import java.sql.SQLException;
+import java.sql.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -59,8 +55,13 @@ public class SystemEngineDatabaseMetaDataTest extends IntegrationTest {
     @Test
     @Tag("v2")
     void getCatalogs() throws SQLException {
-        String database = integration.ConnectionInfo.getInstance().getDatabase();
-        assertTrue(getRows(DatabaseMetaData::getCatalogs).contains(List.of(database)));
+        try (Connection connection = createConnection(); Statement statement = connection.createStatement()) {
+            statement.executeUpdate("CREATE DATABASE test_get_catalogs");
+            String database = integration.ConnectionInfo.getInstance().getDatabase();
+            assertTrue(getRows(DatabaseMetaData::getCatalogs).contains(List.of(database)));
+            assertTrue(getRows(DatabaseMetaData::getCatalogs).contains(List.of("test_get_catalogs")));
+            statement.executeUpdate("DROP DATABASE test_get_catalogs");
+        }
     }
 
     @ParameterizedTest
--- a/src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
+++ b/src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
@@ -26,7 +26,6 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class SystemEngineDatabaseMetaDataTest extends IntegrationTest {
     private static final long ID = ProcessHandle.current().pid() + System.currentTimeMillis();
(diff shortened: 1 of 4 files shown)


src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
```diff
@@ -201,7 +201,15 @@ public ResultSet getTableTypes() throws SQLException {
 
 	@Override
 	public ResultSet getCatalogs() throws SQLException {
-		return createResultSet(Stream.of(entry(TABLE_CAT, TEXT)), List.of(List.of(connection.getCatalog())));
+		List<List<?>> rows = new ArrayList<>();
+		String query = "SELECT CATALOG_NAME AS TABLE_CAT FROM information_schema.catalogs ORDER BY TABLE_CAT";
```
cosmpop, line 205: @bogdantruta-firebolt I suggest to use lower case columns names. By default firebolt as postgress is case insensitive, but this can be changed by config and this may not work

bogdantruta-firebolt: The uppercase column names are also present in the getSchemas methods and all the constants present in MetadataColumns class. I would say to either change only this occurrence of uppercase column names, or maybe investigate/create a task of changing all column name constants to lowercase.

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
```diff
@@ -53,7 +53,14 @@ void readOnly() throws SQLException {
...
-        assertEquals(List.of(List.of("information_schema", database)), getSchemas(DatabaseMetaData::getSchemas));
+        assertEquals(List.of(List.of("information_schema", database)), getRows(DatabaseMetaData::getSchemas));
+    }
+
+    @Test
+    @Tag("v2")
+    void getCatalogs() throws SQLException {
+        String database = integration.ConnectionInfo.getInstance().getDatabase();
+        assertTrue(getRows(DatabaseMetaData::getCatalogs).contains(List.of(database)));
     }
```
ptiurin, line 64: This currently does not test the new functionality. We had the previous getCatalogs return a single database in a list.
A way to improve this would be to create another database (it's cheap to do so in Firebolt) by running sql `CREATE DATABASE test_catalogs` and then check that the result list contains both our current database as well as `test_catalogs`. This way we verify the new behaviour.
Just remember to run `DROP DATABASE test_catalogs` afterwards.

src/main/java/com/firebolt/jdbc/metadata/FireboltDatabaseMetadata.java
cosmpop: scratch that, my bad. to make it upper case the text needs to be wrapped in double quotes "COL_NAME"

cosmpop: But to make it mre clear I think we should use lowercase col names everywherebecause in the execution of the SQL they are llowercaseanyway. @stepansergeevitch WDYT?

cosmpop, line 205: @bogdantruta-firebolt just a question, why did you alias the catalog name to TABLE_CAT. In the ansy standard the name of this column is catalog_name and we return a list of catalog names from this method. why not keep it simple catalog_name?

bogdantruta-firebolt: I saw this done in Postgres' JDBC driver, as well as it being the original implementation

stepansergeevitch (approved):
Generally it looks good to me. I see we have a failing action, let's try to fix it. It should fairly straightforward, we just need to update the action version we use here: https://github.com/firebolt-db/jdbc/blob/master/.github/workflows/code-analysis.yml#L22

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
```diff
@@ -29,6 +25,8 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class SystemEngineDatabaseMetaDataTest extends IntegrationTest {
+    private static final long ID = ProcessHandle.current().pid() + System.currentTimeMillis();
+    private static final String DATABASE_NAME = "jdbc_system_engine_integration_test_" + ID;
```
ptiurin, line 29: Current `database` name is already unique when tests run from the integration tests so we can avoid doing this and just reuse the name + some suffix.

ptiurin (approved):
Otherwise lgtm

PR #498, opened 2025-03-06
FIR-43922: if query_label is set, then respect it on the following statements

goprean:
When using a set statement with a query_label, then the query label should be applied to subsequent statements executed on the same connection or until a new SET query_label is executed. 

Example:
SET query_label = 'Firebolt test1';
SELECT * from my_table;

the select statement should be associated on the backend with the query label specified in the first statement ('Firebolt test1').

Each statement is create a new query label (in the form of a UUID). When sending the query label value to the Firebolt backend check if we don't have the query_label property on the connection properties (meaning is was populated by a SET query_label statement) and use the one from the connection properties rather than the one from the statement itself.


src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -334,6 +328,18 @@ private Map<String, String> getAllParameters(FireboltProperties fireboltProperti
...
+	 * Each statement will have a query label (most of them will be autogenerated except for the SET query_label explicit statement which will use the specified query label).
+	 * When encountering a SET query_label statement we will set it on the firebolt properties and use that one instead the one created on the statementInfoWrapper
+	 * @param fireboltProperties
+	 * @param statementInfoWrapper
+	 * @return
+	 */
+	private String determineQueryLabel(FireboltProperties fireboltProperties, StatementInfoWrapper statementInfoWrapper) {
+		String existingQueryLabel = fireboltProperties.getRuntimeAdditionalProperties().get(FireboltQueryParameterKey.QUERY_LABEL.getKey());
+		return StringUtils.isNotBlank(existingQueryLabel) ? existingQueryLabel : statementInfoWrapper.getLabel();
+	}
```
ptiurin, line 341: Looks like we have several places where we have essentially the same logic. Can we unify this somehow? We're not accessing attributes of this class so this doesn't have to be a method here. We can have it anywhere we want and we can make it static.

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -267,6 +258,96 @@ void setCorrectThenWrongParameter() throws SQLException {
...
 
+	@Test
+	@Tag("v2")
+	void canSetQueryLabelMultipleTimes() throws SQLException {
+		try (Connection connection = createConnection()) {
+			try (Statement statement = connection.createStatement()) {
+				String currentTime = getCurrentTime(statement);
+
+				String firstQueryLabel = "first query label " + RandomStringUtils.randomNumeric(4);
+				assertFalse(statement.execute(String.format("SET query_label = '%s'", firstQueryLabel)));
```
bogdantruta-firebolt, line 269: Although this is correct, I feel the return of the execute method could be confusing due to it being a boolean. I think a better option for a no ResultSet returning execute could be executeUpdate since it throws an error if a ResultSet is returned

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -267,6 +258,96 @@ void setCorrectThenWrongParameter() throws SQLException {
 		setWrongParameter("SET time_zone = 'EST';SET bar=tar", Map.of("time_zone", "EST"), "bar");
 	}
 
+	@Test
+	@Tag("v2")
```
bogdantruta-firebolt, line 262: maybe add a slow tag due to the test having to wait 10 seconds for the query history to update?

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -267,6 +258,96 @@ void setCorrectThenWrongParameter() throws SQLException {
...
+	 * @return
+	 */
+	private void assertQueryFound(Statement statement, String queryLabelValue, String afterTimestamp, String queryText) throws SQLException {
+		String queryHistoryQuery = """
+			SELECT query_text
+			FROM information_schema.engine_query_history
+			WHERE query_label = '%s' and submitted_time > '%s' and query_text = '%s';
+			""";
+		ResultSet resultSet = statement.executeQuery(String.format(queryHistoryQuery, queryLabelValue, afterTimestamp, queryText));
+		assertTrue(resultSet.next(), () -> "Did not find query with the specified query label");
```
bogdantruta-firebolt, line 329: Here you use a supplier function, while in the next method you use a plain string. I think a plain String could be enough

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -267,6 +258,96 @@ void setCorrectThenWrongParameter() throws SQLException {
...
+	}
+
+	/**
+	 * Runs the query: select now() and retuns the result
+	 */
+	private String getCurrentTime(Statement statement) throws SQLException {
+		ResultSet resultSet = statement.executeQuery("SELECT now()");
+		resultSet.next();
+		return resultSet.getString(1);
+	}
```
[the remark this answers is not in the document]
goprean: I will change it then to always assume that the server side is storing the dates as UTC.

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -267,6 +258,96 @@ void setCorrectThenWrongParameter() throws SQLException {
...
+		}
+	}
+
+	private void sleepForMillis(long millis) {
+        try {
+            Thread.sleep(millis);
+        } catch (InterruptedException e) {
+            // do nothing
+        }
+    }
```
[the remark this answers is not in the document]
goprean: Hmm... in my IDE is showing the same as the other methods. How does it show for you?

src/test/java/com/firebolt/jdbc/client/query/StatementClientImplTest.java
```diff
@@ -107,6 +93,51 @@ private Entry<String, String> shouldPostSqlQuery(boolean systemEngine) throws SQ
...
+		String actualQuery = getActualRequestString(actualRequest);
+		assertEquals("show databases;", actualQuery);
+
+		HttpUrl httpUrl = actualRequest.url();
+		assertEquals("http", httpUrl.scheme());
+		assertEquals("firebolt1", httpUrl.host());
+		assertEquals(555, httpUrl.port());
+
+		Set<String> queryParameterNames= httpUrl.queryParameterNames();
+		assertEquals(queryParameterNames.size(), 3);
```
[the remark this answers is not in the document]
goprean: Will remove the rest of the assertion and will only keep the check for the query_label parameter.

.legitignore
```diff
@@ -1,2 +1,3 @@
 instance_51B6E36F85D19C9CA81AECD8DE760121 # type: secret, commit: 0c27c6d, file: src/test/java/com/firebolt/jdbc/client/query/StatementClientImplTest.java, line: 161
-instance_2A43FC8CB9D52E7BBCFBE224988CE371 # type: secret, commit: 0c27c6d, file: src/test/java/com/firebolt/jdbc/client/query/StatementClientImplTest.java, line: 162
\ No newline at end of file
+instance_2A43FC8CB9D52E7BBCFBE224988CE371 # type: secret, commit: 0c27c6d, file: src/test/java/com/firebolt/jdbc/client/query/StatementClientImplTest.java, line: 162
```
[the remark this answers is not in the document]
goprean: I don't really undestand how this scanning tool is working. From what I could tell the problem was that we are reading the auth token in code. Since I am not checking the request for these values I might need to change this file at all.

src/integrationTest/java/integration/tests/StatementTest.java
goprean: I have removed the assertion as this was not what was being tested for. I just wanted to make sure that the statement was executed correctly. I added the test before checking the code and I thought that the SET query_label statement is not supported but it is. So this assertion is not needed any more.

src/integrationTest/java/integration/tests/StatementTest.java
goprean: Done.

src/integrationTest/java/integration/tests/StatementTest.java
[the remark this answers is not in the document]
ptiurin: Looks messed up in github, but fine in IDE. I think I know what's going on here - tabs vs spaces. Everywhere else there are tabs and this method has spaces.
I have no preference, but let's keep it consistent.

src/test/java/com/firebolt/jdbc/client/query/StatementClientImplTest.java
```diff
@@ -107,6 +93,43 @@ private Entry<String, String> shouldPostSqlQuery(boolean systemEngine) throws SQ
...
+		StatementInfoWrapper statementInfoWrapper = StatementUtil.parseToStatementInfoWrappers("show databases").get(0);
+		assertNotNull(statementInfoWrapper.getLabel());
+		statementClient.executeSqlStatement(statementInfoWrapper, fireboltProperties, fireboltProperties.isSystemEngine(), QUERY_TIMEOUT);
+
+		verify(okHttpClient).newCall(requestArgumentCaptor.capture());
+		Request actualRequest = requestArgumentCaptor.getValue();
+		Map<String, String> expectedHeaders = new LinkedHashMap<>();
+		expectedHeaders.put("Authorization", "Bearer token"); // legit:ignore
+		expectedHeaders.put("User-Agent", userAgent("ConnB/2.0.9 JDBC/%s (Java %s; %s %s; ) ConnA/1.0.9"));
+		assertEquals(expectedHeaders, extractHeadersMap(actualRequest));
```
ptiurin, line 120: I'd remove this block as well - user agent is non-essential part of our code and I really would like to avoid having more tests than necessary to depend on it.

ptiurin (commented):
Minor comment, otherwise looks good.
Please also have a look at the sonar [issues](https://sonarcloud.io/project/issues?id=firebolt-db_jdbc&pullRequest=498&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true). They're generally pretty basic but it's good to have them fixed before merging.

goprean:
> Minor comment, otherwise looks good. Please also have a look at the sonar [issues](https://sonarcloud.io/project/issues?id=firebolt-db_jdbc&pullRequest=498&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true). They're generally pretty basic but it's good to have them fixed before merging.

My bad, in my previous company the Sonar check would fail if there were errors. The quality gate was configured in such a way to fail. I did not even see that there were failures but I have addressed them

ptiurin:
> My bad, in my previous company the Sonar check would fail if there were errors. The quality gate was configured in such a way to fail. I did not even see that there were failures but I have addressed them

Yeah, not sure why we don't fail for those issues. We have a default quality gate set up, but we can look into extending it.

ptiurin (approved):
lgtm

PR #505, opened 2025-03-07
feat: FIR-44116 Line and Column information missing from JDBC errors

bogdantruta-firebolt:
Updated ServerError creation to not trim the error message anymore, but to leave it complete. Now the error message is 1:1 with the one received from the server, so no more missing the line/column information


ptiurin (approved):
lgtm, let's make sure all integration tests (including v1) pass before merging.

goprean (approved):

PR #509, opened 2025-03-10
feat: FIR-38161 Fix unauthorized error message fetching in ServerError

bogdantruta-firebolt:
When a statement is executed from a client which is blocked / not allowed via network policy, the error message from the exception is null, due to the message being found via another key ("detail")

### DIFF
--- a/src/integrationTest/java/integration/tests/ConnectionTest.java
+++ b/src/integrationTest/java/integration/tests/ConnectionTest.java
@@ -243,7 +243,7 @@ class ConnectionTest extends IntegrationTest {
 
     @Test
     @Tag("v2")
-    void networkPolicyBlockedServiceAccountThrowsErrorWhenFetchingEngineUrl() throws SQLException {
+    void networkPolicyBlockedServiceAccountThrowsError() throws SQLException {
         try (Connection connection = createConnection();
              Statement statement = connection.createStatement()) {
             //this is done as to not clutter the environment variables for one test
--- a/src/integrationTest/java/integration/tests/ConnectionTest.java
+++ b/src/integrationTest/java/integration/tests/ConnectionTest.java
@@ -27,28 +27,59 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 @TestInstance(TestInstance.Lifecycle.PER_METHOD)
 class ConnectionTest extends IntegrationTest {
 
     enum EngineType {
-        SYSTEM_ENGINE, CUSTOM_ENGINE
+        CUSTOM_ENGINE, SYSTEM_ENGINE
     }
 
     @ParameterizedTest(name = "{0}")
     @Tag("v2")
     @EnumSource(EngineType.class)
-    void connectToNotExistingDb(EngineType engineType) {
+    void connectToNotExistingDbV2(EngineType engineType) throws SQLException {
         String database = "wrong_db";
-        assumeTrue(EngineType.CUSTOM_ENGINE.equals(engineType));
-        ConnectionInfo params = integration.ConnectionInfo.getInstance();
-        String engineSuffix = EngineType.CUSTOM_ENGINE.equals(engineType) ? "&engine=" + params.getEngine() : "";
-        String url = format("jdbc:firebolt:%s?env=%s&account=%s%s", database, params.getEnv(), params.getAccount(), engineSuffix);
-        FireboltException e = assertThrows(FireboltException.class, () -> DriverManager.getConnection(url, params.getPrincipal(), params.getSecret()));
+        boolean useCustomEngine = EngineType.CUSTOM_ENGINE.equals(engineType);
+        ConnectionInfo params = ConnectionInfo.getInstance();
+        String url = getConnectionInfoWithEngineAndDatabase(useCustomEngine ? params.getEngine() : null, database).toJdbcUrl();
+        FireboltException e;
+        if (useCustomEngine) {
+            e = assertThrows(FireboltException.class, () -> DriverManager.getConnection(url, params.getPrincipal(), params.getSecret()));
+        } else {
+            // the connection will be successful because v2 does not set database or engine on system engine connection
+            try (Connection connection = DriverManager.getConnection(url, params.getPrincipal(), params.getSecret());
+                ResultSet resultSet = connection.createStatement().executeQuery("SELECT 1")) {
+                assertTrue(resultSet.next());
+                assertEquals(1, resultSet.getInt(1));
+                //even on system engine, the database should be set, but since database does not exist, it should throw an error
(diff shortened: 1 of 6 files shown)


src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -220,6 +221,52 @@ void preparedStatementBatchesWorkIfMergeParameterProvided() throws SQLException
         }
     }
 
+	@Test
+	@Tag("v2")
+	@Tag("slow")
```
goprean, line 226: What makes this slow? Just the fact that we create the user/service and network policy?

src/main/java/com/firebolt/jdbc/exception/ServerError.java
```diff
@@ -126,12 +136,12 @@ public Error(String code, String name, Severity severity, Source source, String
             this.location = location;
         }
 
-        Error(JSONObject json) {
+        Error(JSONObject json, boolean isError) {
```
goprean, line 139: Maybe it's just me but it reads a bit funky to create an Error object and then to pass in a flag to specify if it is an error.

src/integrationTest/java/integration/tests/ConnectionTest.java
bogdantruta-firebolt: For some reason, the delete user query takes around 6-7 seconds, so all in this test takes around 10 seconds I would say, that's why I added the tag

src/main/java/com/firebolt/jdbc/exception/ServerError.java
```diff
@@ -29,12 +28,23 @@ public ServerError(Query query, Error[] errors) {
...
-        return jsonArray == null ? null : IntStream.range(0, jsonArray.length()).boxed().map(jsonArray::getJSONObject).map(factory).toArray(arrayFactory);
-    }
+	private static Error[] fromJson(JSONArray jsonArray) {
+		return jsonArray == null
+                ? null
+				: IntStream.range(0, jsonArray.length()).boxed()
+                        .map(jsonArray::getJSONObject)
+                        .map(jsonObject -> new Error(jsonObject, true))
+						.toArray(Error[]::new);
+	}
```
ptiurin, line 47: There are some tabs here as well. Let's clean them up so we have everything following the same standard. Please check the other files too.

src/main/java/com/firebolt/jdbc/exception/ServerError.java
```diff
@@ -46,7 +56,7 @@ public String getErrorMessage() {
                 :
                 Arrays.stream(errors)
                         .filter(Objects::nonNull)
-                        .map(e -> Stream.of(e.severity, e.source, e.code, e.name, e.description).filter(Objects::nonNull).map(Object::toString).collect(joining(" ")))
+                        .map(e -> Stream.of(e.severity, e.code, e.name, e.description).filter(Objects::nonNull).map(Object::toString).collect(joining(" ")))
```
ptiurin, line 59: Why did we remove `e.source` here?

src/main/java/com/firebolt/jdbc/exception/ServerError.java
```diff
@@ -126,12 +136,12 @@ public Error(String code, String name, Severity severity, Source source, String
...
         }
 
-        Error(JSONObject json) {
+        Error(JSONObject json, boolean isError) {
             this(json.optString("code", null),
                     json.optString("name", null),
                     json.optEnum(Severity.class, "severity"),
                     ofNullable(json.optString("source", null)).map(Source::fromText).orElse(Source.UNKNOWN),
-                    json.optString("description", null),
+                    json.optString(isError ? "description" : "detail", null),
```
ptiurin, line 144: We can probably eliminate the need for `isError` altogether by using something like `json.optString("description", json.optString("detail", null))`

src/main/java/com/firebolt/jdbc/exception/ServerError.java
bogdantruta-firebolt: It seems to always be set to Unknown, and to just show that in the error message didn't make sense

ptiurin: Let's not remove things arbitrarily. Source is a part of an [error spec](https://docs.google.com/document/d/1MBWqDWxrfSiBQNbHf7CCqGmVPZORNmkPZh9P_6tpFtE/edit?tab=t.0#heading=h.gfozuvdf3bwx) from the backend, we should track it if it's provided.
We should use a similar filter to `.filter(Objects::nonNull)` but for unknown if you want to improve this if "Unknown" is coming from us. If it's coming from the backend - let them drive it.

src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -256,6 +241,46 @@ void preparedStatementBatchesWorkIfMergeParameterProvided() throws SQLException
...
+    void networkPolicyBlockedServiceAccountThrowsError() throws SQLException {
+        try (Connection connection = createConnection();
+             Statement statement = connection.createStatement()) {
+            //this is done as to not clutter the environment variables for one test
+            ConnectionInfo connectionInfo = getConnectionInfoForNetworkPolicyTest(statement);
+            String jdbcUrl = getJdbcUrl(connectionInfo, true, true);
+
+            //this should fail when executing setting the database
+            FireboltException exception = assertThrows(FireboltException.class,
+                    () -> DriverManager.getConnection(jdbcUrl, connectionInfo.getPrincipal(), connectionInfo.getSecret()));
```
goprean, line 255: FireboltDriver fireboltDriver = new FireboltDriver();
fireboltDriver.connect(url, properties)

since that's how a client would interact with the driver.

goprean (approved):

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -127,7 +129,7 @@ void shouldThrowExceptionWhenExecutingWrongQueryV1() throws SQLException {
 
 	@Test
 	@Tag("v2")
-	@EnvironmentCondition(value = "4.2.0", comparison = EnvironmentCondition.Comparison.GE)
+	@EnvironmentCondition(value = "4.2.0", attribute = fireboltVersion, comparison = EnvironmentCondition.Comparison.GE)
```
ptiurin, line 132: Why is this changing?

ptiurin (approved):
Apart from that looks good

src/integrationTest/java/integration/tests/StatementTest.java
bogdantruta-firebolt: This was by default comparing if the connection is v1 or v2 and now it should check the PackDb version if I remember correctly

ptiurin: Let's double check, because `fireboltVersion` can be v1 or v2 and we're comparing it to `4.2.0`? This doesn't sound right.

src/integrationTest/java/integration/tests/ConnectionTest.java
bogdantruta-firebolt: I haven't seen any integration tests use this, but the DriverManager.getConnection does kind of the same thing, searching for suitable drivers. Do you think we should change this?

goprean: Never mind. My bad. Actually the way the client would interact with the driver is DriverManager.getConnection(). All good.

PR #515, opened 2025-03-14
feat: FIR-44260 - do not validate connection when changing the paramters set from the server

goprean:
Some API calls to the server return response headers that we need to set on the connection. 

When setting those connection properties, we are currently validating the connection by making a SELECT 1 statement call. (Same happens for reset call)

There are 2 problems:
 1. we should trust the parameter headers that come from the server and should not do additional validation when setting the property
 2. when we the server mentions that the connection needs to be reset we are clearing the properties and also validate the connection. The problem is that we do the SELECT 1 on the system engine url and not on the user engine url.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
+++ b/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
@@ -40,6 +40,8 @@ import okhttp3.internal.http2.StreamResetException;
 @CustomLog
 public class StatementClientImpl extends FireboltClient implements StatementClient {
 
+	private static final boolean DO_NOT_VALIDATE_CONNECTION = false;
+
 	private static final String TAB_SEPARATED_WITH_NAMES_AND_TYPES_FORMAT = "TabSeparatedWithNamesAndTypes";
 	private static final Map<Pattern, String> missConfigurationErrorMessages = Map.of(
 			Pattern.compile("HTTP status code: 401"), "Please associate user with your service account.",
@@ -348,7 +350,7 @@ public class StatementClientImpl extends FireboltClient implements StatementClie
 			}
 			for (String header : response.headers(HEADER_UPDATE_PARAMETER)) {
 				String[] keyValue = header.split("=");
-				connection.addProperty(keyValue[0].trim(), keyValue[1].trim());
+				connection.addProperty(keyValue[0].trim(), keyValue[1].trim(), DO_NOT_VALIDATE_CONNECTION);
 			}
 		}
 	}
--- a/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
+++ b/src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
@@ -1,25 +1,5 @@
 package com.firebolt.jdbc.client.query;
 
-import static com.firebolt.jdbc.connection.settings.FireboltQueryParameterKey.*;
-import static com.firebolt.jdbc.exception.ExceptionType.INVALID_REQUEST;
-import static com.firebolt.jdbc.exception.ExceptionType.UNAUTHORIZED;
-import static java.lang.String.format;
-import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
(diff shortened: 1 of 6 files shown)


src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -106,6 +109,66 @@ void shouldNotFetchTokenNorEngineHostForLocalFirebolt() throws SQLException {
...
+    @Test
+    void canAddPropertyThatWillValidateConnection() throws SQLException {
+        try (FireboltConnection connection = createConnection(SYSTEM_ENGINE_URL, connectionProperties)) {
+            // part of the connection there are some calls to the firebolt statements, so reset the mocks before calling reset
+            Mockito.reset(fireboltStatementService);
+
+            // by default, it will validate the connection
+            connection.addProperty("newProperty", "new value");
+
+            // verify no calls are made on the connection
```
bogdantruta-firebolt, line 147: I think this comment should say that call are made

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -106,6 +109,66 @@ void shouldNotFetchTokenNorEngineHostForLocalFirebolt() throws SQLException {
...
+            verify(fireboltStatementService).execute(any(), any(FireboltProperties.class), any(FireboltStatement.class));
+
+            // initial additional property should still be there
```
bogdantruta-firebolt, line 150: this I think should also say that the property should be there since there was no call to methods that could change/ remove it

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -106,6 +109,66 @@ void shouldNotFetchTokenNorEngineHostForLocalFirebolt() throws SQLException {
...
+            // part of the connection there are some calls to the firebolt statements, so reset the mocks before calling reset
+            Mockito.reset(fireboltStatementService);
+
+            // do not validate the connection
+            connection.addProperty("newProperty", "new value", false);
+
+            // verify no calls are made on the connection
+            verify(fireboltStatementService, never()).execute(any(), any(FireboltProperties.class), any(FireboltStatement.class));
+
+            // initial additional property should still be there
```
bogdantruta-firebolt, line 167: same as above

goprean: Here no calls are made so the comment seems fine, no?

bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
```diff
@@ -40,6 +40,8 @@
 @CustomLog
 public class StatementClientImpl extends FireboltClient implements StatementClient {
 
+	private static final boolean DO_NOT_VALIDATE_CONNECTION = false;
```
stepansergeevitch, line 43: Ok this got me confused initially. I like the idea of this variable for readability, but I got initially confused thinking it's the actual setting for the class - like DO_NOT_VALIDATE_CONNECTION==false -> we do validate a connection. I would suggest to improve the naming to avoid confusion. Maybe add a `_FLAG` in the end? Or maybe there is a better naming scheme

stepansergeevitch (approved):
Generally LGTM, one comment

src/main/java/com/firebolt/jdbc/client/query/StatementClientImpl.java
goprean: For me this comes down to code readability where the parameter is actually used:
for example to me this
connection.addProperty(keyValue[0].trim(), keyValue[1].trim(), DO_NOT_VALIDATE_CONNECTION);

reads better than this:
connection.addProperty(keyValue[0].trim(), keyValue[1].trim(), false);

As from the code review I do not know what false means. But the variable name makes it a bit clearer.

PR #516, opened 2025-03-14
Fir 44260 reduce the number of requests during the connection establishing part2

goprean:
Part2 
- cache the response for the use database and use engine calls
- for now we know what are the parameters that the server sends back so we keep those results in cache and update those values when we get the connection from cache. We also need to update the value for the endpoint after we do a use engine since next calls should use the user engine.

### DIFF
--- a/src/integrationTest/java/integration/tests/CachedConnectionTest.java
+++ b/src/integrationTest/java/integration/tests/CachedConnectionTest.java
@@ -9,6 +9,8 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class CachedConnectionTest extends IntegrationTest {
@@ -18,18 +20,20 @@ public class CachedConnectionTest extends IntegrationTest {
 
     @BeforeAll
     void beforeAll() {
-        executeStatementFromFile("/statements/cache-connection/ddl.sql");
+        executeStatementFromFile("/statements/cached-connection/ddl.sql");
     }
 
     @AfterAll
     void afterEach() {
-        executeStatementFromFile("/statements/cache-connection/cleanup.sql");
+        executeStatementFromFile("/statements/cached-connection/cleanup.sql");
     }
 
     @Test
     @Tag("v2")
     @Tag("slow")
     void createTwoConnections() throws SQLException {
+        String testStartTime = getCurrentUTCTime();
+
         // create a connection on the first engine and database
         try (Connection connection = createConnection()) {
             ResultSet rs = connection.createStatement().executeQuery("SELECT 101");
@@ -47,13 +51,53 @@ public class CachedConnectionTest extends IntegrationTest {
             ResultSet rs = connection.createStatement().executeQuery("SELECT 103");
             assertTrue(rs.next());
 
-            sle
+            // wait for the query history to propagate
+            sleepForMillis(10000);
+
+            // check that the connection cached was using the correct connection to the db and engine
+            String queryHistoryQueryFormat = """
+				SELECT query_text
+				FROM information_schema.engine_query_history
+				WHERE submitted_time > '%s' and status = 'STARTED_EXECUTION' and (query_text='SELECT 101;' or query_text='SELECT 103;')
+				order by submitted_time desc
+			""";
+            ResultSet engineOneResultSet = connection.createStatement().executeQuery(String.format(queryHistoryQueryFormat, testStartTime));
+            assertTrue(engineOneResultSet.next());
+            assertEquals("SELECT 103;", engineOneResultSet.getString(1));
+
+            assertTrue(engineOneResultSet.next());
+            assertEquals("SELECT 101;", engineOneResultSet.getString(1));
+
+            assertFalse(engineOneResultSet.next());
+
         }
 
         // create a connection on the second engine and database
         try (Connection connection = createConnection(CACHED_TEST_ENGINE_NAME, CACHED_TEST_DATABASE_NAME)) {
             ResultSet rs = connection.createStatement().executeQuery("SELECT 104");
             assertTrue(rs.next());
+
+            // wait for the query history to propagate
+            sleepForMillis(10000);
+
+            // check that the connection cached was using the correct connection to the db and engine
+            String queryHistoryQueryFormat = """
+				SELECT query_text
+				FROM information_schema.engine_query_history
+				WHERE submitted_time > '%s' and status = 'STARTED_EXECUTION' and (query_text='SELECT 102;' or query_text='SELECT 104;')
+				order by submitted_time desc
+			""";
+
+            ResultSet engineTwoResultSet = connection.createStatement().executeQuery(String.format(queryHistoryQueryFormat, testStartTime));
+            assertTrue(engineTwoResultSet.next());
+            assertEquals("SELECT 104;", engineTwoResultSet.getString(1));
+
+            assertTrue(engineTwoResultSet.next());
+            assertEquals("SELECT 102;", engineTwoResultSet.getString(1));
+
+            assertFalse(engineTwoResultSet.next());
+
+
         }
 
 
--- a/src/integrationTest/java/integration/tests/CachedConnectionTest.java
+++ b/src/integrationTest/java/integration/tests/CachedConnectionTest.java
@@ -1,0 +1,62 @@
+package integration.tests;
+
+import integration.IntegrationTest;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class CachedConnectionTest extends IntegrationTest {
+
+    private static final String CACHED_TEST_ENGINE_NAME = "cached_test_second_engine";
+    private static final String CACHED_TEST_DATABASE_NAME = "cached_test_second_db";
+
+    @BeforeAll
+    void beforeAll() {
+        executeStatementFromFile("/statements/cache-connection/ddl.sql");
+    }
+
+    @AfterAll
+    void afterEach() {
+        executeStatementFromFile("/statements/cache-connection/cleanup.sql");
+    }
+
+    @Test
+    @Tag("v2")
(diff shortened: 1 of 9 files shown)


src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,127 @@
...
-            statement.executeUpdate(use("ENGINE", properties.getEngine()));
+            verifyEngineExists(statement, properties.getHost(), properties.getEngine());
         }
         // now session properties are updated with new database and engine
         FireboltProperties sessionProperties = fireboltConnection.getSessionProperties();
         return new Engine(fireboltConnection.getEndpoint(), null, sessionProperties.getEngine(), sessionProperties.getDatabase(), null);
     }
 
+    /**
+     * We are caching the database names for some time and if we had checked it before assumed that it is still valid.
```
bogdantruta-firebolt, line 59: i think the description here sounds a little weird

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,127 @@
...
+            // as of Mar 2025 we know that as a side effect of calling "use database <xxx>" we are updating the database parameter on the connection.
+            // so if we want to use the value from cache we need to save this value on the connection when we use the cached connection
+            List<Pair<String, String>> cachedValuesForDatabase = List.of(
+                    Pair.of("database", fireboltConnection.getSessionProperties().getDatabase()));
+            CACHED_VERIFIED_DATABASES.put(asCacheKey(host, databaseName), cachedValuesForDatabase, cacheDatabaseDurationInSeconds, SECONDS);
+        }
+    }
+
+    /**
+     * We are caching the database names for 15 mins and if we had checked it before assumed that it is still valid.
```
bogdantruta-firebolt, line 88: This I think should indicate a cache on engine names and also it should not provide a time value (or at least set the time value to 1 hour)

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,127 @@
...
         }
         // now session properties are updated with new database and engine
         FireboltProperties sessionProperties = fireboltConnection.getSessionProperties();
         return new Engine(fireboltConnection.getEndpoint(), null, sessionProperties.getEngine(), sessionProperties.getDatabase(), null);
     }
 
+    /**
+     * We are caching the database names for some time and if we had checked it before assumed that it is still valid.
+     *
+     * @param databaseName
```
bogdantruta-firebolt, line 61: param is not up to date

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,127 @@
...
+            List<Pair<String, String>> cachedValuesForDatabase = List.of(
+                    Pair.of("database", fireboltConnection.getSessionProperties().getDatabase()));
+            CACHED_VERIFIED_DATABASES.put(asCacheKey(host, databaseName), cachedValuesForDatabase, cacheDatabaseDurationInSeconds, SECONDS);
+        }
+    }
+
+    /**
+     * We are caching the database names for 15 mins and if we had checked it before assumed that it is still valid.
+     *
+     * @param engineName
```
bogdantruta-firebolt, line 90: param here is not up to date

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -38,12 +54,200 @@ void getEndine(String database) throws SQLException {
...
+        when(statementFromFirstConnection.executeUpdate("USE DATABASE " + MY_DATABASE)).thenReturn(1);
+        when(statementFromFirstConnection.executeUpdate("USE ENGINE " + MY_ENGINE)).thenReturn(1);
+        when(firstConnection.getSessionProperties()).thenReturn(properties);
+        when(firstConnection.getEndpoint()).thenReturn(MY_ENGINE_ENDPOINT);
+
+        FireboltEngineVersion2Service service = new FireboltEngineVersion2Service(firstConnection);
+        Engine actualEngine = service.getEngine(properties);
+        assertEquals(new Engine(MY_ENGINE_ENDPOINT, null, MY_ENGINE, MY_DATABASE, null), actualEngine);
+
+        // should not execute any caching
```
bogdantruta-firebolt, line 90: I think this is slightly confusing. Something like: "Should run use statements since cache is empty" would be better.

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -38,12 +54,200 @@ void getEndine(String database) throws SQLException {
...
+        Statement statementFromSecondConnection = mock(Statement.class);
+        when(secondConnection.createStatement()).thenReturn(statementFromSecondConnection);
+        when(secondConnection.getSessionProperties()).thenReturn(secondConnectionProperties);
+        when(secondConnection.getEndpoint()).thenReturn(MY_ENGINE_ENDPOINT);
+
+        doNothing().when(secondConnection).addProperty(anyString(), anyString());
+        doNothing().when(secondConnection).setEndpoint(anyString());
+
+        service = new FireboltEngineVersion2Service(secondConnection);
+        Engine actualEngineFromCache = service.getEngine(secondConnectionProperties);
```
bogdantruta-firebolt, line 113: I think a comment above this stating that getEngine should use cache information would help

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -38,12 +54,200 @@ void getEndine(String database) throws SQLException {
...
+        when(secondConnection.getSessionProperties()).thenReturn(secondConnectionProperties);
+        when(secondConnection.getEndpoint()).thenReturn(MY_ENGINE_ENDPOINT);
+        when(statementFromSecondConnection.executeUpdate("USE DATABASE " + MY_DATABASE)).thenReturn(1);
+        when(statementFromSecondConnection.executeUpdate("USE ENGINE " + MY_ENGINE)).thenReturn(1);
+
+        doNothing().when(secondConnection).addProperty(anyString(), anyString());
+        doNothing().when(secondConnection).setEndpoint(anyString());
+
+        service = new FireboltEngineVersion2Service(secondConnection);
+        Engine actualEngineFromCache = service.getEngine(secondConnectionProperties);
```
bogdantruta-firebolt, line 177: I think this engine is not actually from cache :))))

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -38,12 +54,200 @@ void getEndine(String database) throws SQLException {
...
+        assertEquals(new Engine(MY_ENGINE_ENDPOINT, null, MY_ENGINE, MY_DATABASE, null), actualEngineFromCache);
+
+        // should not execute any caching
```
bogdantruta-firebolt, line 180: same as previous test

src/integrationTest/java/integration/tests/CachedConnectionTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+            // wait for the query history to propagate. Checking the query history will make sure that the queries were executed on the same engine
+            sleepForMillis(TimeUnit.SECONDS.toMillis(10));
+
+            assertQueryFoundInHistory(connection.createStatement(), testStartTime, "INSERT INTO statement_test_cached(id) values (102);");
+            assertQueryFoundInHistory(connection.createStatement(), testStartTime, "INSERT INTO statement_test_cached(id) values (104);");
+        }
+
+    }
+    private void assertQueryFoundInHistory(Statement statement, String afterTimestamp, String queryText) throws SQLException {
```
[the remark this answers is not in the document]
goprean: I will take a look at how many I do have and will try to move these methods to the base class as more tests can use it.

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,127 @@
...
+     */
+    private void verifyDatabaseExists(Statement statement, String host, String databaseName) throws SQLException {
+        synchronized (CACHED_VERIFIED_DATABASES) {
+
+            if (CACHED_VERIFIED_DATABASES.containsKey(asCacheKey(host, databaseName))) {
+                log.debug("Using cache verification of database");
+
+                // need to set the values on the connection that were cached on the original use database call
+                for (Pair<String, String> pair : CACHED_VERIFIED_DATABASES.get(asCacheKey(host, databaseName))) {
+                    fireboltConnection.addProperty(pair.getKey(), pair.getValue());
```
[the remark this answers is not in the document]
goprean: Great catch. Thanks! I need to also check why there was no test failing as we should have a test that fails.

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
goprean: After reading it again, yeah, it sounds weird. In my mind it made sense. I had re-word it.

bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
```diff
@@ -3,34 +3,138 @@
...
+    /**
+     * Once we verify a database with the backend we will cache its result and would consider the database valid in between connection establishing.
+     * If the database was deleted between the time a connection cached the database and the second connection using the cached value, then this would only
+     * be caught while executing the statement, not during connection time.
+     *
+     * @param statement - statement that will execute the verification of the database
+     * @param host - the system engine url
+     * @param databaseName - the name of the database to check
+     */
+    private void verifyDatabaseExists(Statement statement, String host, String databaseName) throws SQLException {
```
stepansergeevitch, line 69: I believe the naming is slightly confusing. We're not only verifying the database exists, but most importantly we set the database property on the `fireboltConnection`. I would suggest to improve the naming, say something like `setDatabaseProperty` (and similarly `setEngineProperty`), just an example

stepansergeevitch (approved):
LGTM, one small comment

src/main/java/com/firebolt/jdbc/service/FireboltEngineVersion2Service.java
goprean: I agree. A lot of these methods have side effects which is not reflected in the method names.

PR #523, opened 2025-03-21
FIR-43922: refactored a bit the connection creation

goprean:
Part2 - refactored the creating of a FireboltConnection. 
We will create 3 types of connections based on the url and the connection params:
 - FireboltConnectionUserPassword (existing)
 - FireboltConnectionServiceSecret (existing)
 - FireboltLocalhostConnectionServiceSecret (new) - for testing the case of localhost

This will give us a chance to validate parameters that we know are mandatory. For example for clientId/clientSecret connection we need to make sure the clientId, clientSecret, accountName exists (non-null values) since we will use that as a key for the connection cache

### DIFF
--- a/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionProvider.java
+++ b/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionProvider.java
@@ -1,0 +1,87 @@
+package com.firebolt.jdbc.connection;
+
+import com.firebolt.jdbc.connection.settings.FireboltProperties;
+import com.firebolt.jdbc.type.ParserVersion;
+import com.firebolt.jdbc.util.PropertyUtil;
+import java.sql.SQLException;
+import java.util.Properties;
+import java.util.regex.Pattern;
+import lombok.NonNull;
+
+import static java.lang.String.format;
+
+/**
+ * Based on the connection url and the connection properties this class will generate the correct firebolt connection
+ */
+public class FireboltConnectionProvider {
+
+
+    private FireboltConnectionProviderWrapper fireboltConnectionProviderWrapper;
+
+    public FireboltConnectionProvider() {
+        this(new FireboltConnectionProviderWrapper());
+    }
+
+    // visible for testing
+    FireboltConnectionProvider(FireboltConnectionProviderWrapper fireboltConnectionProviderWrapper) {
+        this.fireboltConnectionProviderWrapper = fireboltConnectionProviderWrapper;
+    }
+
+    /**
+     * From the url and connection properties, determine the appropriate firebolt connection
+     */
+    public FireboltConnection create(@NonNull String url, Properties connectionSettings) throws SQLException {
+        switch(getUrlVersion(url, connectionSettings)) {
+            case 1:
+                return fireboltConnectionProviderWrapper.createFireboltConnectionUsernamePassword(url, connectionSettings, ParserVersion.LEGACY);
+            case 2:
+                return connectToLocalHost(url, connectionSettings) ?
+                        fireboltConnectionProviderWrapper.createLocalhostFireboltConnectionServiceSecret(url, connectionSettings, ParserVersion.CURRENT) :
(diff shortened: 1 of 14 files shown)


goprean:
Integration tests also passed: 
 v1- https://github.com/firebolt-db/jdbc/actions/runs/13992305058
 v2 - https://github.com/firebolt-db/jdbc/actions/runs/13992308123

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionProvider.java
```diff
@@ -0,0 +1,87 @@
...
+
+
+    private FireboltConnectionProviderWrapper fireboltConnectionProviderWrapper;
+
+    public FireboltConnectionProvider() {
+        this(new FireboltConnectionProviderWrapper());
+    }
+
+    // visible for testing
+    FireboltConnectionProvider(FireboltConnectionProviderWrapper fireboltConnectionProviderWrapper) {
```
bogdantruta-firebolt, line 26: This could be changed for a static factory class, and testing could use mockStatic like here: https://www.baeldung.com/mockito-mock-static-methods

goprean: I'be been giving our discussion a bit more thought and I would still prefer to use the constructor as a pattern rather than mock static methods for 2 reasons:
 - cleaner tests as we would create the mocks an injected into the test object (this can be done as you pointed out using mock statics)
 - clear separation of concerns. The FireboltConnectionProvider knows how to check the url and detect what connection to create, and the FireboltConnectionProviderWrapper (which probably a better name would be FireboltConnectionFactory) knows how to create each of the connection. Also if we want to pass in a different FireboltConnectionFactory implementation we don't have to change the FireboltConnectionProvider. 

What do you think @bogdantruta-firebolt?

bogdantruta-firebolt: I like this approach, I'm thinking you can go for it!

bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionUserPassword.java
```diff
@@ -57,6 +58,25 @@ protected void authenticate() throws SQLException {
         sessionProperties = loginProperties.toBuilder().host(engine.getEndpoint()).engine(engine.getName()).database(database).build();
     }
 
+    @Override
+    protected void validateConnectionParameters() throws SQLException {
+        String clientId = loginProperties.getPrincipal();
```
stepansergeevitch, line 63: nit: Since it's a username/password connection, let's rename clientId->username and clientSecret -> password for clarity

src/main/java/com/firebolt/jdbc/connection/LocalhostFireboltConnectionServiceSecret.java
```diff
@@ -0,0 +1,72 @@
...
+import java.sql.SQLException;
+import java.util.Optional;
+import java.util.Properties;
+import lombok.NonNull;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * A Connection to firebolt that is using localhost as the url of the firebolt server. It will talk to a firebolt 2.0 server.
+ */
+public class LocalhostFireboltConnectionServiceSecret extends FireboltConnectionServiceSecret {
```
stepansergeevitch, line 19: Is it actually ServiceSecret, as I understand we're always passing accessToken for this connection. I believe it should be enough to call it `LocalhostFireboltConnection`

stepansergeevitch (approved):
A couple of comments for naming

src/main/java/com/firebolt/jdbc/connection/LocalhostFireboltConnectionServiceSecret.java
goprean: I did not know if there was a localhost connection for v1 that needs to be supported for local testing, but Mosha confirmed that v1 is not used for testing. Forgot to rename it.

PR #524, opened 2025-03-24
FIR-44620: add connection cache to url parameters

goprean:
Add a new connection parameter: cache_connection

This will default to true for clientId/clientSecret connections (both for actual backend and when using localhost connections).

If the parameter is used on a username/password connection (Firebolt 1.0 backend) it will be ignored but a warning will be displayed to the user.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
+++ b/src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
@@ -160,17 +160,10 @@ public abstract class FireboltConnection extends JdbcBase implements Connection,
 	protected abstract void validateConnectionParameters() throws SQLException;
 
 	/**
-	 * By default, the connection caching is supported only on the clientId/clientSecret connections.
-	 * @return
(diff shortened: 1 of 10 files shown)


bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -159,6 +159,19 @@ protected void connect() throws SQLException {
 	 */
 	protected abstract void validateConnectionParameters() throws SQLException;
 
+	/**
+	 * By default, the connection caching is supported only on the clientId/clientSecret connections.
+	 * @return
+	 */
+	protected boolean isConnectionCachingEnabled() {
```
stepansergeevitch, line 166: Should this method be abstract, and the implementation moved to the UserPassword child?

goprean: I was thinking that the default is false and then whichever connection wants to enable it, it will override the method as needed. If you think it will make it clearer I will make it abstract and let each connection implementation provide the default.

PR #525, opened 2025-03-24
FIR-44260: service connection can only connect to infra 2 version

goprean:
At some point the clientId/clientSecret combination was supported for version 1 infra. This is now only supported for infra version 2+. 
So now when we create a connection we already know the infra version. Added an abstract method to get the infraVersion on the connection base class and provide the implementation for each of the connections: username/password and clientId/clientSecret.


goprean:
Integration tests also passed for this branch: https://github.com/firebolt-db/jdbc/actions/runs/14040471153

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -54,14 +54,6 @@ void shouldNotGetEngineUrlOrDefaultEngineUrlWhenUsingSystemEngine() throws SQLEx
...
     }
 
-    @Test
-    void notExistingDb() throws SQLException {
-        connectionProperties.put("database", "my_db");
-        when(fireboltGatewayUrlService.getUrl(any(), any())).thenReturn("http://my_endpoint");
-        when(fireboltEngineService.doesDatabaseExist("my_db")).thenReturn(false);
-        assertEquals("Database my_db does not exist", assertThrows(FireboltException.class, () -> createConnection("jdbc:firebolt:?env=dev&account=dev", connectionProperties)).getMessage());
-    }
-
```
ptiurin, line 64: Why are we removing this test?

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -151,6 +150,11 @@ protected void connect() throws SQLException {
 		log.debug("Connection opened");
 	}
 
+	/**
+	 * Returns the version of the firebolt backend the connection is established to
+	 */
+	public abstract int getInfraVersion();
+
```
ptiurin, line 157: Do we use it anywhere in our code? If not, we can probably remove infra version altogether. Firebolt 2.0 used to have account v1 and v2 that had different metadata capabilities. Since we only have accounts v2 now there's no need to keep track of the version.

stepansergeevitch (approved):
LGTM from my end, please resolve remaining comments

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
ptiurin: Ye, I get that this is a base. I was wondering if we use this method anywhere else, but I see a couple of places where we use it to determine if we're running on Firebolt 2.0 or 1.0.

ptiurin (approved):

PR #526, opened 2025-03-25
FIR-44260: addressed comments from previous review

goprean:
Addressed comments open on this review: https://github.com/firebolt-db/jdbc/pull/523
as I have closed the PR and then I saw the comments.


goprean:
Added some more tests as now the Sonar is seeing all of this as new code and thus failing the new code test coverage threshold

PR #528, opened 2025-03-26
feat: FIR-44260 add connection caching

goprean:
Cache the connection (only the in memory caching is provided in the review). The on disk cache will be sent as a separate review. 

CacheService will be created by the CacheProvider which will be a static instance. This means it will have the same cacheService across multiple invocations (a.k.a - multiple connections creations).

ConnectionCache is the object that will the object that will be stored in cache. When this object is created the first time we will store the connectionId of the connection that created it and saved it in cache. 

DatabaseOptions is a container for saving the database information for a database name. Currently we only save the db name but if more info will be needed in the future will be adding to this object.

EngineOptions is a container for saving engine information for an engine name. We currently store just the engine endpoint and the name of the engine but if more info will be needed in the future will be adding to this object.

Additionally we will cache the access token and the system engine url. 

NOTE: 
1. We are still caching the token and the system engine url downstream but I will be removing this code in subsequent review. 
2. If this review is too big (and I know it is) I can try to break it up in smaller reviews but will need some more time.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/cache/InMemoryCacheService.java
+++ b/src/main/java/com/firebolt/jdbc/cache/InMemoryCacheService.java
@@ -1,0 +1,40 @@
+package com.firebolt.jdbc.cache;
+
+import com.firebolt.jdbc.cache.exception.CacheException;
+import com.firebolt.jdbc.cache.key.CacheKey;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import net.jodah.expiringmap.ExpirationPolicy;
+import net.jodah.expiringmap.ExpiringMap;
+
+/**
+ * Cache service that uses the memory to store the cached values.
+ *
+ * NOTE: this should be package protected as the clients should use the {@link CacheServiceProvider} to create a cache service.
+ */
+class InMemoryCacheService implements CacheService {
+
+    // by default cache the connection for 1hr
+    private static final int DEFAULT_CACHE_TTL_IN_HOURS = 1;
+
+    private ExpiringMap<String, ConnectionCache> map;
+
+    public InMemoryCacheService() {
+        this(ExpiringMap.builder().variableExpiration().build());
+    }
+
+    // visible for testing
+    InMemoryCacheService(ExpiringMap<String, ConnectionCache> map) {
+        this.map = map;
+    }
+
+    @Override
+    public void put(CacheKey key, ConnectionCache connectionCache) throws CacheException {
+        map.put(key.getValue(), connectionCache, ExpirationPolicy.CREATED, DEFAULT_CACHE_TTL_IN_HOURS, TimeUnit.HOURS);
+    }
+
+    @Override
+    public Optional<ConnectionCache> get(CacheKey key) throws CacheException {
+        return Optional.ofNullable(map.get(key.getValue()));
+    }
+}
--- a/src/main/java/com/firebolt/jdbc/cache/key/CacheKey.java
+++ b/src/main/java/com/firebolt/jdbc/cache/key/CacheKey.java
@@ -2,7 +2,7 @@ package com.firebolt.jdbc.cache.key;
 
 
 /**
- * Implementations should make sure that the CacheKey has equals and hash methods implemented
+ * The interface of a caching key.
  */
 public interface CacheKey {
 
--- a/src/main/java/com/firebolt/jdbc/cache/key/CacheKey.java
+++ b/src/main/java/com/firebolt/jdbc/cache/key/CacheKey.java
@@ -1,0 +1,14 @@
+package com.firebolt.jdbc.cache.key;
+
+
+/**
+ * Implementations should make sure that the CacheKey has equals and hash methods implemented
+ */
+public interface CacheKey {
+
+    /**
+     * Returns the cache key value
+     * @return
+     */
+    String getValue();
+}
--- a/src/main/java/com/firebolt/jdbc/cache/key/ClientSecretCacheKey.java
+++ b/src/main/java/com/firebolt/jdbc/cache/key/ClientSecretCacheKey.java
@@ -1,0 +1,24 @@
+package com.firebolt.jdbc.cache.key;
+
+import lombok.CustomLog;
+import lombok.EqualsAndHashCode;
+
+@CustomLog
+@EqualsAndHashCode
+public class ClientSecretCacheKey implements CacheKey {
+
+    private String value;
+
+    public ClientSecretCacheKey(String clientId, String clientSecret, String accountName) {
+        value = hashValues(clientId, clientSecret, accountName);
+    }
+
+    @Override
+    public String getValue() {
+        return value;
+    }
+
+    private String hashValues(String clientId, String clientSecret, String accountName) {
+        return String.join("#", clientId, clientSecret, accountName);
+    }
+}
--- a/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
+++ b/src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
@@ -1,6 +1,10 @@
 package com.firebolt.jdbc.connection;
 
 import com.firebolt.jdbc.annotation.ExcludeFromJacocoGeneratedReport;
+import com.firebolt.jdbc.cache.CacheService;
+import com.firebolt.jdbc.cache.ConnectionCache;
+import com.firebolt.jdbc.cache.key.CacheKey;
+import com.firebolt.jdbc.cache.key.ClientSecretCacheKey;
 import com.firebolt.jdbc.client.account.FireboltAccountRetriever;
 import com.firebolt.jdbc.client.authentication.AuthenticationRequest;
 import com.firebolt.jdbc.client.authentication.FireboltAuthenticationClient;
@@ -9,8 +13,6 @@ import com.firebolt.jdbc.client.gateway.GatewayUrlResponse;
 import com.firebolt.jdbc.connection.settings.FireboltProperties;
 import com.firebolt.jdbc.exception.FireboltException;
 import com.firebolt.jdbc.service.FireboltAuthenticationService;
-import com.firebolt.jdbc.service.FireboltEngineInformationSchemaService;
-import com.firebolt.jdbc.service.FireboltEngineService;
 import com.firebolt.jdbc.service.FireboltEngineVersion2Service;
 import com.firebolt.jdbc.service.FireboltGatewayUrlService;
 import com.firebolt.jdbc.service.FireboltStatementService;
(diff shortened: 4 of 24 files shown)


ptiurin:
Will review this later, but let's keep the ticket in the title of form `FIR-XXXXX`. This enables jira linking from GH.

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
```diff
@@ -21,52 +23,137 @@
...
+            return getAccessToken(loginProperties).orElse("");
+        } else {
+            // caching is enabled so check the cache first
+            String accessTokenFromCache = connectionCache.getAccessToken();
+            if (StringUtils.isNotBlank(accessTokenFromCache)) {
+                return accessTokenFromCache;
+            }
+
+            // not in cache so get it from the source.
+            synchronized (connectionCache) {
```
bogdantruta-firebolt, line 118: I think this should have been the cacheService maybe?

bogdantruta-firebolt, line 118: I get a suggestion from IntelliJ to make the fields used in synchronized final

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -212,8 +265,152 @@ void willSetConnectionCacheAsFalseForAnyValueNotBooleanTrue() throws SQLExceptio
...
+
+    @Test
+    void willGetJwtAndSystemEngineUrlFromCacheWhenConnectionCachingIsEnabledAndTheValuesAreInCache() throws SQLException {
+        enableCacheConnection();
+
+        // cache is present
+        ConnectionCache connectionCache = new ConnectionCache(A_CONNECTION_ID);
+        connectionCache.setAccessToken(AN_ACCESS_TOKEN);
+        connectionCache.setSystemEngineUrl(SYSTEM_ENGINE_URL_FOR_DEV_ACCOUNT);
+        lenient().when(mockCacheService.get(cacheKey)).thenReturn(Optional.of(connectionCache));
```
bogdantruta-firebolt, line 319: I don't think this should be lenient either

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -212,8 +265,152 @@ void willSetConnectionCacheAsFalseForAnyValueNotBooleanTrue() throws SQLExceptio
...
+        }
+    }
+
+    @Test
+    void canCacheJwtTokenSystemEngineDatabaseAndUserEngineWhenConnectionIsCachable() throws SQLException {
+        enableCacheConnection();
+
+        // no cache is present for the key
+        lenient().when(mockCacheService.get(cacheKey)).thenReturn(Optional.empty());
+        lenient().doNothing().when(mockCacheService).put(eq(cacheKey), any(ConnectionCache.class));
```
bogdantruta-firebolt, line 346: this two as well

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -212,8 +265,152 @@ void willSetConnectionCacheAsFalseForAnyValueNotBooleanTrue() throws SQLExceptio
...
+            assertNotNull(connectionCache.getConnectionId());
+            assertEquals(AN_ACCESS_TOKEN, connectionCache.getAccessToken());
+            assertEquals(SYSTEM_ENGINE_URL_FOR_DEV_ACCOUNT, connectionCache.getSystemEngineUrl());
+
+            // the engine options and db options are set as a side effect of this method so all we can check is make sure it is called
+            verify(fireboltEngineVersion2Service).getEngine(any(), any());
+
+            FireboltProperties sessionProperties = fireboltConnection.getSessionProperties();
+            assertFalse(sessionProperties.isSystemEngine());
+            assertTrue(sessionProperties.isCompress());
```
bogdantruta-firebolt, line 365: this should also probably be imported from jupier.api

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -212,8 +265,152 @@ void willSetConnectionCacheAsFalseForAnyValueNotBooleanTrue() throws SQLExceptio
...
+
+        // cache is present
+        ConnectionCache connectionCache = new ConnectionCache(A_CONNECTION_ID);
+        connectionCache.setAccessToken(AN_ACCESS_TOKEN);
+        connectionCache.setSystemEngineUrl(SYSTEM_ENGINE_URL_FOR_DEV_ACCOUNT);
+        connectionCache.setDatabaseOptions(DB_NAME, new DatabaseOptions(List.of(Pair.of("database", DB_NAME))));
+        EngineOptions engineOptions = new EngineOptions(USER_ENGINE_ENDPOINT, List.of(Pair.of("engine", ENGINE_NAME)));
+        connectionCache.setEngineOptions(ENGINE_NAME, engineOptions);
+
+        lenient().when(mockCacheService.get(cacheKey)).thenReturn(Optional.of(connectionCache));
```
bogdantruta-firebolt, line 385: lenient here too

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -45,210 +79,197 @@ void getEngine(String database) throws SQLException {
...
-        FireboltProperties secondConnectionProperties = new FireboltProperties(secondProps);
+        verify(mockFireboltStatement).executeUpdate("USE DATABASE \"" + MY_DATABASE + "\"");
+        verify(mockFireboltStatement).executeUpdate("USE ENGINE \"" + MY_ENGINE + "\"");
+    }
 
-        FireboltConnection secondConnection = mock(FireboltConnection.class);
-        when(secondConnection.getSessionProperties()).thenReturn(secondConnectionProperties);
+    @Test
+    void canGetDatabaseAndEngineFromSourceWhenCachingIsEnabledButDatabaseNotInCache() throws SQLException {
+        when(mockConnectionCache.getDatabaseOptions(MY_DATABASE)).thenReturn(Optional.empty());
```
bogdantruta-firebolt, line 118: this is duplicated

src/test/java/com/firebolt/jdbc/service/FireboltEngineVersion2ServiceTest.java
```diff
@@ -45,210 +79,197 @@ void getEngine(String database) throws SQLException {
...
-        // sleep for 2 seconds just to make sure the cache expired
-        sleepForMillis(TimeUnit.SECONDS.toMillis(2));
-
-        Properties secondProps = new Properties();
-        secondProps.setProperty("database", MY_DATABASE);
-        secondProps.setProperty("engine", MY_ENGINE);
-        secondProps.setProperty("host", host);
+        FireboltEngineVersion2Service service = new FireboltEngineVersion2Service(mockFireboltConnection);
+        Engine actualEngine = service.getEngine(properties, Optional.of(mockConnectionCache));
+        assertEquals(new Engine(MY_ENGINE_ENDPOINT, null, MY_ENGINE, MY_DATABASE, null), actualEngine);
```
bogdantruta-firebolt, line 176: this should also be imported from jupiter api

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
goprean: The cache object is created only once but I cannot have it as final since it is not created directly in the constructor

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
goprean: One checks that there are no values in cache and one mocks a successful save on the cache service. So both are needed.

src/main/java/com/firebolt/jdbc/cache/key/CacheKey.java
```diff
@@ -0,0 +1,14 @@
...
+ * Implementations should make sure that the CacheKey has equals and hash methods implemented
+ */
+public interface CacheKey {
+
+    /**
+     * Returns the cache key value
+     * @return
+     */
+    String getValue();
+}
```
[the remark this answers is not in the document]
goprean: I will remove the comment from the class. Since we are not converting the key to a String, and that is used as a key to a map, we don't need to rely on the subclasses to implement hashing and equals.

src/main/java/com/firebolt/jdbc/cache/key/ClientSecretCacheKey.java
```diff
@@ -0,0 +1,24 @@
...
+        value = hashValues(clientId, clientSecret, accountName);
+    }
+
+    @Override
+    public String getValue() {
+        return value;
+    }
+
+    private String hashValues(String clientId, String clientSecret, String accountName) {
+        return String.join("#", clientId, clientSecret, accountName);
```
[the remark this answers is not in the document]
goprean: We could have joined the values without any delimiter. I had thought If we want to print the key in logs, we could parse the string by  the "#" and then replace the clientSecret with "****" (basically obfuscate it). But we could still see what accounts and what clientId was being used.

src/main/java/com/firebolt/jdbc/cache/InMemoryCacheService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public void put(CacheKey key, ConnectionCache connectionCache) throws CacheException {
+        map.put(key.getValue(), connectionCache, ExpirationPolicy.CREATED, DEFAULT_CACHE_TTL_IN_HOURS, TimeUnit.HOURS);
+    }
+
+    @Override
+    public Optional<ConnectionCache> get(CacheKey key) throws CacheException {
+        return Optional.ofNullable(map.get(key.getValue()));
+    }
```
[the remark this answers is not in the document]
goprean: The class should take care and not return the expired key. I have added a new unit test to take care of this scenario. I have not used this class but have used other libraries that offered the same functionality.

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
```diff
@@ -19,51 +21,139 @@
...
+     * If caching is enabled, then this value will contain the cached values
+     */
+    private ConnectionCache connectionCache;
+
+    /**
+     * Each connection will have its unique id
+     */
+    private String connectionId;
+
+    @SuppressWarnings("java:S107") // ignore the too many params. Max is 7. Refactor this method at some point
```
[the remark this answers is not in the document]
goprean: I have removed the parserVersion as this will always be current for the clientId/clientSecret connections as this is the v2 version of fireboat

ptiurin (approved):
Apart from that looks good

src/test/java/com/firebolt/jdbc/cache/InMemoryCacheServiceTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertTrue(connectionCacheOptional.isEmpty());
+
+        inMemoryCacheService.put(cacheKey, mockConnectionCache);
+
+        connectionCacheOptional = inMemoryCacheService.get(cacheKey);
+        assertSame(connectionCacheOptional.get(), mockConnectionCache);
+    }
+
+    @Test
+    @SuppressWarnings("java:S2925")
```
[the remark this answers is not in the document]
goprean: Done.

PR #529, opened 2025-03-26
feat: FIR-42858 Add server side async query support

bogdantruta-firebolt:


src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+            String token = statement.getAsyncToken();
+
+            assertNotNull(token);
+            assertFalse(token.isEmpty());
+            assertTrue(connection.isAsyncQueryRunning(token));
+
+            sleepForMillis(TimeUnit.SECONDS.toMillis(5));
+
+            assertFalse(connection.isAsyncQueryRunning(token));
+            assertTrue(connection.isAsyncQuerySuccessful(token));
```
goprean, line 52: Can we use the same token to get the result multiple times? what would happen if we use the token twice. Let's add some asserts statements.

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+            try (ResultSet resultSet = statement.executeQuery("SELECT count(*) FROM async_table_test")) {
+                assertTrue(resultSet.next());
+                assertEquals(1, resultSet.getInt(1));
```
goprean, line 56: what are we testing here? That we can also make sync queries to the table?

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+    void executeCancelServerSideAsyncQueryTest() throws SQLException {
+        try (FireboltConnection connection = createConnection().unwrap(FireboltConnection.class);
+             FireboltStatement statement = connection.createStatement().unwrap(FireboltStatement.class)) {
+            statement.executeAsync("INSERT INTO async_table_test SELECT checksum(*) FROM GENERATE_SERIES(1, 2500000000)");
+            String token = statement.getAsyncToken();
+
+            assertNotNull(token);
+            assertFalse(token.isEmpty());
+
+            sleepForMillis(TimeUnit.SECONDS.toMillis(1));
```
goprean, line 72: this sleep does not seem to be needed right since we cancel the query

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+            try (ResultSet resultSet = statement.executeQuery("SELECT count(*) FROM async_table_test")) {
+                assertTrue(resultSet.next());
+                assertEquals(1, resultSet.getInt(1));
+            }
+        }
+    }
+
+    @Test
+    @Tag("v2")
+    void executeCancelServerSideAsyncQueryTest() throws SQLException {
```
goprean, line 63: Let's add one more tests that test the following: 
- would cancel the query after the query has completed. 
- cannot use an invalid token to get the result or cancel the query
- is token per engine? it would be interesting to change the engine and database and try to use the token for a different engine and db?
- lets add calls to make that non-insert queries are not supported and check the error messages as these are going to be seen by the clients.

src/main/java/com/firebolt/jdbc/client/query/StatementClient.java
```diff
@@ -12,7 +12,7 @@ public interface  StatementClient {
 	 * Post SQL statement
 	 */
 	InputStream executeSqlStatement(StatementInfoWrapper statementInfoWrapper, FireboltProperties connectionProperties,
-									boolean systemEngine, int queryTimeout) throws SQLException;
+									boolean systemEngine, int queryTimeout, boolean isServerAsync) throws SQLException;
```
goprean, line 15: if the query is executed asynchronous do we still need the queryTimeout. I see that the queryTimeout is set: 
			if (queryTimeout > 0) {
				params.put("max_execution_time", String.valueOf(queryTimeout));
			}

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +686,47 @@ public void register(CacheListener listener) {
...
 		cacheListeners.forEach(CacheListener::cleanup);
 	}
+
+	public boolean isAsyncQueryRunning(String asyncQueryToken) throws SQLException {
+        return getAsyncQueryStatus(asyncQueryToken).equals("RUNNING");
+	}
+
+	public boolean isAsyncQuerySuccessful(String asyncQueryToken) throws SQLException {
+        return getAsyncQueryStatus(asyncQueryToken).equals("ENDED_SUCCESSFULLY");
+	}
```
goprean, line 696: Can you check the indentation here? I seems a bit off

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +686,47 @@ public void register(CacheListener listener) {
...
+
+	public boolean cancelAsyncQuery(String asyncQueryToken) throws SQLException {
+		if (asyncQueryToken == null || asyncQueryToken.isEmpty()) {
```
goprean, line 699: you can use StringUtils.isBlank(asyncQueryToken)

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +686,47 @@ public void register(CacheListener listener) {
...
+		try (PreparedStatement statement = createPreparedStatement("CALL fb_GetAsyncStatus(?)")) {
+			statement.setString(1, asyncQueryToken);
+			try (ResultSet rs = statement.executeQuery()) {
+				rs.next();
+				asyncQueryId = rs.getString("query_id");
+			}
+		} catch (SQLException ex) {
+			throw new FireboltException("Could not check the status of the async query", ex);
+		}
+		if (asyncQueryId != null && !asyncQueryId.isEmpty()) {
```
goprean, line 712: same here: use StringUtils.isNotBlank(asyncQueryId)

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +686,47 @@ public void register(CacheListener listener) {
...
+			try (PreparedStatement statement = createPreparedStatement("CANCEL QUERY WHERE query_id = ?")) {
+				statement.setString(1, asyncQueryId);
+				statement.executeUpdate();
+			}
+		}
+		return true;
```
goprean, line 718: this does not seem to be right when cannot find the asyncQueryId. In case this does not exist are we suppose to return true? seems misleading to the client that the query was canceled. To me it seems a better response would be to return false. or throw an exception that would let them know that no query is associated with that token.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +686,47 @@ public void register(CacheListener listener) {
...
+	}
+
+	private String getAsyncQueryStatus(String asyncQueryToken) throws SQLException {
+		try (PreparedStatement statement = prepareStatement("CALL fb_GetAsyncStatus(?)")) {
+			statement.setString(1, asyncQueryToken);
+			try (ResultSet rs = statement.executeQuery()) {
+				rs.next();
+				return rs.getString("status");
```
goprean, line 726: this would throw an exception if there is no row. It would be better to check if the next()returns false and throw a proper exception.

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -115,27 +115,30 @@ private Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper) t
...
+                setOrAppendFirstUnclosedStatementResult(statementInfoWrapper, resultSet);
+            }
 		} else {
 			log.warn("Aborted query with id {}", determineQueryLabel(statementInfoWrapper));
 		}
 		return Optional.ofNullable(resultSet);
 	}
 
-	private String determineQueryLabel(StatementInfoWrapper statementInfoWrapper) {
+    private void setOrAppendFirstUnclosedStatementResult(StatementInfoWrapper statementInfoWrapper, ResultSet resultSet) {
```
goprean, line 133: I see that this call was protected by synchronized in the original call, and not protected here. Any reason?

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -203,7 +206,45 @@ protected int executeUpdate(List<StatementInfoWrapper> sql) throws SQLException
...
+                        runningStatementLabel), ex);
+                throw ex;
+            } finally {
+                runningStatementLabel = null;
+                synchronized (statementsToExecuteLabels) {
+                    statementsToExecuteLabels.remove(query.getLabel());
+                }
+            }
+            synchronized (this) {
+                setOrAppendFirstUnclosedStatementResult(query, null);
```
goprean, line 240: are we passing null as the result set? Would that not cause any NPEs down the line? Can you double check?

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
```diff
@@ -29,7 +29,7 @@
 import static java.lang.String.format;
 
 public class FireboltConnectionServiceSecret extends FireboltConnection {
-    private static final String PROTOCOL_VERSION = "2.1";
+    private static final String PROTOCOL_VERSION = "2.3";
```
goprean, line 32: Can you double check if 2.3 or 2.4 that supports async? I thought it was 2.4

src/test/java/com/firebolt/jdbc/service/FireboltStatementServiceTest.java
```diff
@@ -19,15 +19,14 @@
...
 import java.sql.SQLException;
 
 import static java.util.Optional.empty;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
goprean, line 29: In my IDE I have setup not to collapse for less than 100. We should decide on what to do so we don't have these import changes. I prefer to see what classes are imported from which package rather than not knowing what classes are actually used. So my preference would be to keep this expanded not collapsed with *

src/test/java/com/firebolt/jdbc/service/FireboltStatementServiceTest.java
```diff
@@ -106,7 +105,51 @@ void shouldBeEmptyWithNonQueryStatement() throws SQLException {
...
+	void shouldExecuteQueryAsyncWithEmptyResponse() throws SQLException {
+		StatementInfoWrapper statementInfoWrapper = StatementUtil
+				.parseToStatementInfoWrappers("INSERT INTO ltv SELECT * FROM ltv_external").get(0);
+		FireboltProperties fireboltProperties = fireboltProperties("localhost", false);
+
+		FireboltStatementService fireboltStatementService = new FireboltStatementService(statementClient);
+		FireboltStatement statement = mock(FireboltStatement.class);
+		when(statement.getQueryTimeout()).thenReturn(10);
+		when(statementClient.executeSqlStatement(statementInfoWrapper, fireboltProperties, false, 10, true)).thenReturn(getMockInputStream(""));
+		assertThrows(FireboltException.class, () -> fireboltStatementService.executeAsyncStatement(statementInfoWrapper, fireboltProperties, statement));
```
goprean, line 137: Let's also validate the error message here since this message will most likely be seen by the end user .

src/test/java/com/firebolt/jdbc/service/FireboltStatementServiceTest.java
```diff
@@ -106,7 +105,51 @@ void shouldBeEmptyWithNonQueryStatement() throws SQLException {
...
+		when(statement.getQueryTimeout()).thenReturn(10);
+		String jsonResult = "{\"token\":\"123\"}";
+		when(statementClient.executeSqlStatement(statementInfoWrapper, fireboltProperties, false, 10, true)).thenReturn(getMockInputStream(jsonResult));
+		String asyncToken = fireboltStatementService.executeAsyncStatement(statementInfoWrapper, fireboltProperties, statement);
+		assertEquals("123", asyncToken);
+		verify(statementClient).executeSqlStatement(statementInfoWrapper, fireboltProperties, false, 10, true);
+	}
+
+	@Test
+	void shouldExecuteQueryAsyncWithEmptyResponse() throws SQLException {
```
goprean, line 128: the name of the method implies that everything should be fine, where actually it throws an exception. So maybe the naming of the method should be re-written

src/test/java/com/firebolt/jdbc/service/FireboltStatementServiceTest.java
```diff
@@ -106,7 +105,51 @@ void shouldBeEmptyWithNonQueryStatement() throws SQLException {
...
+		FireboltStatementService fireboltStatementService = new FireboltStatementService(statementClient);
+		FireboltStatement statement = mock(FireboltStatement.class);
+		when(statement.getQueryTimeout()).thenReturn(10);
+		when(statementClient.executeSqlStatement(statementInfoWrapper, fireboltProperties, false, 10, true)).thenReturn(getMockInputStream(""));
+		assertThrows(FireboltException.class, () -> fireboltStatementService.executeAsyncStatement(statementInfoWrapper, fireboltProperties, statement));
+		verify(statementClient).executeSqlStatement(statementInfoWrapper, fireboltProperties, false, 10, true);
+	}
+
+	@Test
+	void shouldExecuteQueryAsyncWithNullResponse() throws SQLException {
```
goprean, line 142: same comments as above. Rename the method name to imply that is not successful and verify the error message

src/test/java/com/firebolt/jdbc/statement/FireboltStatementTest.java
```diff
@@ -599,4 +599,43 @@ void shouldNotOutputAWSKeyIdInDebugLogs(String query) throws SQLException {
...
+        when(fireboltStatementService.executeAsyncStatement(any(), any(), any())).thenReturn(asyncToken);
+        fireboltStatement.executeAsync("INSERT INTO x values (1)");
+        assertEquals(asyncToken, fireboltStatement.getAsyncToken());
+        assertEquals(0, fireboltStatement.getUpdateCount());
+        assertFalse(fireboltStatement.getMoreResults());
+        assertNull(fireboltStatement.getResultSet());
+    }
+
+    @Test
+    void shouldExecuteAsyncStatementAndThrowFireboltException() throws SQLException {
```
goprean, line 618: the name of the test does not specify what is wrong that an exception is being thrown. The test body implies that we cannot read the response from the server.

src/test/java/com/firebolt/jdbc/statement/FireboltStatementTest.java
```diff
@@ -599,4 +599,43 @@ void shouldNotOutputAWSKeyIdInDebugLogs(String query) throws SQLException {
...
+        FireboltConnection connection = mock(FireboltConnection.class);
+        when(connection.getSessionProperties()).thenReturn(fireboltProperties);
+        FireboltStatement fireboltStatement = new FireboltStatement(fireboltStatementService, fireboltProperties, connection);
+        when(fireboltStatementService.executeAsyncStatement(any(), any(), any())).thenThrow(new FireboltException("Cannot read response from DB"));
+        assertThrows(FireboltException.class, () -> fireboltStatement.executeAsync("INSERT 1"));
+    }
+
+    @Test
+    void shouldExecuteAsyncStatementAndThrowFireboltExceptionDueToQueryTypeBeingSelect() {
```
goprean, line 627: nit: maybe something like cannotExecuteSelectStatementsAsSelect. But this brings me to another question. Why not support select statements. Those can take up quite some time. I though that only the SET statements we cannot execute async.

src/test/java/com/firebolt/jdbc/statement/FireboltStatementTest.java
```diff
@@ -599,4 +599,43 @@ void shouldNotOutputAWSKeyIdInDebugLogs(String query) throws SQLException {
...
+        FireboltConnection connection = mock(FireboltConnection.class);
+        when(connection.getSessionProperties()).thenReturn(fireboltProperties);
+        FireboltStatement fireboltStatement = new FireboltStatement(fireboltStatementService, fireboltProperties, connection);
+        assertThrows(FireboltException.class, () -> fireboltStatement.executeAsync("SELECT 1"));
+    }
+
+    @Test
+    void shouldExecuteAsyncStatementAndThrowFireboltExceptionDueToQueryTypeBeingSet() {
```
goprean, line 635: nit: cannotExecuteSetStatementsAsync

src/integrationTest/java/integration/tests/AsyncQueryTest.java
bogdantruta-firebolt: We test that when the query status function returns that the async query ended successfully, we make sure that the insert actually finished and the rows are inserted, we don't check the value, but we check the row count

src/integrationTest/java/integration/tests/AsyncQueryTest.java
bogdantruta-firebolt: this sleep was intended for the query to actually start, I got it from Petro's tests from the .NET implementation. Do you think we should remove this sleep?

src/main/java/com/firebolt/jdbc/client/query/StatementClient.java
ptiurin: I think it makes sense to have it. The user might want to make sure a query does not keep the engine on for too long, since that incurs cost. If you do COPY from s3, you might not know how much data you have there currently so this can be another failsafe.

ptiurin:
Looks like we have a mix of tabs and spaces again in this change.

src/integrationTest/java/integration/tests/AsyncQueryTest.java
ptiurin: I think I had it beacause metadata might have a small update delay after scheduling a query. If we try to fetch the status straight away the query might not be runnig yet.

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -203,7 +206,45 @@ protected int executeUpdate(List<StatementInfoWrapper> sql) throws SQLException
...
+        if (isStatementNotCancelled(query)) {
+            runningStatementLabel = determineQueryLabel(query);
+            synchronized (this) {
+                validateStatementIsNotClosed();
+            }
+            try {
+                log.debug("Executing the statement with label {} : {}", query.getLabel(),
+                        sanitizeSql(query.getSql()));
+                if (query.getType() != StatementType.NON_QUERY) {
+                    throw new FireboltException("Only non-queries are supported for async statements");
```
ptiurin, line 224: Technically "SET x=y" is also a non-query. Maybe let's rephrase this to make sure it's clear what types of queries are supported.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
bogdantruta-firebolt: It's due to spaces versus tabs indentation. We should align ourselves on what we should use.

src/main/java/com/firebolt/jdbc/service/FireboltStatementService.java
```diff
@@ -52,6 +60,29 @@ public Optional<ResultSet> execute(StatementInfoWrapper statementInfoWrapper,
...
+			if (properties.isCompress()) {
+				inputStreamReader = new InputStreamReader(new LZ4InputStream(is), UTF_8);
+			} else {
+				inputStreamReader = new InputStreamReader(is, UTF_8);
+			}
+			BufferedReader bufferedReader = new BufferedReader(inputStreamReader, properties.getBufferSize());
+			return new JSONObject(bufferedReader.lines().reduce(String::concat).orElseThrow(IOException::new)).optString("token");
+        } catch (IOException e) {
+			throw new FireboltException("Cannot read response from DB: error while reading asyncToken ", e);
+        } catch (JSONException e) {
```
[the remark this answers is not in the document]
bogdantruta-firebolt: The input stream requires the IOException catch, creation of JSONObject does not require the catch branch, but it helps prevent any unwanted empty responses. Maybe we could talk about this tomorrow to see a better alternative.

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
bogdantruta-firebolt: indentation issue, it actually is caught in synchronized

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -203,7 +206,45 @@ protected int executeUpdate(List<StatementInfoWrapper> sql) throws SQLException
...
 	}
 
-	@Override
+	public void executeAsync(String sql) throws SQLException {
+		StatementInfoWrapper query = StatementUtil.parseToStatementInfoWrappers(sql).get(0);
+        createValidator(query.getInitialStatement(), connection).validate(query.getInitialStatement());
+        synchronized (statementsToExecuteLabels) {
+            statementsToExecuteLabels.add(query.getLabel());
+        }
+        if (isStatementNotCancelled(query)) {
```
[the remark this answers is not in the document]
bogdantruta-firebolt: I thought about following the execution plan of a normal query and I did not want to modify this logic. I also thought about this. Should I remove the currentRunning steps?

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
```diff
@@ -203,7 +206,45 @@ protected int executeUpdate(List<StatementInfoWrapper> sql) throws SQLException
...
+                validateStatementIsNotClosed();
+            }
+            try {
+                log.debug("Executing the statement with label {} : {}", query.getLabel(),
+                        sanitizeSql(query.getSql()));
+                if (query.getType() != StatementType.NON_QUERY) {
+                    throw new FireboltException("Only non-queries are supported for async statements");
+                }
+                asyncToken = statementService.executeAsyncStatement(query, sessionProperties, this);
+                currentUpdateCount = 0;
```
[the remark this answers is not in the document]
bogdantruta-firebolt: This comes from the execution of queries vs non queries and I set the value to the nonQuery one since only non queries can be ran async

src/main/java/com/firebolt/jdbc/statement/FireboltStatement.java
bogdantruta-firebolt: This shouldn't be an issue since ResultSet is set to null already for non-queries

src/main/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecret.java
bogdantruta-firebolt: It's 2.3 here, I found out by mistakenly trying to run queries on protocol 2.1 :)))

src/test/java/com/firebolt/jdbc/service/FireboltStatementServiceTest.java
bogdantruta-firebolt: I'll make this change in my IDE too and I'll optimize imports for this.

src/test/java/com/firebolt/jdbc/statement/FireboltStatementTest.java
bogdantruta-firebolt: momentarily only non queries can run async. either way, we would need to modify the drivers too when backend implements async select queries

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    @Tag("v2")
+    void executeServerSideAsyncTest() throws SQLException {
+        try (FireboltConnection connection = createConnection().unwrap(FireboltConnection.class);
+             FireboltStatement statement = connection.createStatement().unwrap(FireboltStatement.class)) {
+            statement.executeAsync("INSERT INTO async_table_test SELECT checksum(*) FROM GENERATE_SERIES(1, 2500000000)");
+            String token = statement.getAsyncToken();
+
+            assertNotNull(token);
+            assertFalse(token.isEmpty());
```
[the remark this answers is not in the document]
goprean: How about assertTrue(StringUtils.isNotEmpty(token)). For me double negations seem harder to read so I try avoiding them as much as possible.

src/integrationTest/java/integration/tests/AsyncQueryTest.java
goprean: If we add more tests that would insert in the same table then this test would fail. Should we get the count before the  async query and then get the count after the async query so then this test will not fail if we add more tests in this class.

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,207 @@
...
+            assertTrue(connection.isAsyncQueryRunning(token));
+
+            statement.execute(format("USE ENGINE %s", engineName));
+            assertEquals(connection.getSessionProperties().getEngine(), engineName);
+
+            assertTrue(connection.isAsyncQueryRunning(token));
+
+            assertTrue(connection.isAsyncQueryRunning(token));
```
goprean, line 158: This can be removed. you only need to check once if the query is running

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,207 @@
...
+            FireboltException exception = assertThrows(FireboltException.class, () -> connection.cancelAsyncQuery(token));
+            assertTrue(exception.getMessage().contains("Attempt to cancel query with query_id"));
+
+            assertTrue(connection.isAsyncQueryRunning(token));
+
+            try (ResultSet resultSet = statement.executeQuery("SELECT count(*) FROM async_table_test")) {
+                assertTrue(resultSet.next());
+                assertEquals(0, resultSet.getInt(1));
+            }
+        }
```
goprean, line 168: I don't get this part. The name of the test says that you can cancel the query on different engine, yet the cancel call throws an exception and the query is still running. (line 162). Should we wait until the query is not running to make sure we don't have any results?

src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+            assertNotNull(token);
+            assertFalse(token.isEmpty());
+
+            sleepForMillis(TimeUnit.SECONDS.toMillis(1));
+
+            assertFalse(connection.isAsyncQueryRunning(token));
+
+            FireboltException exception = assertThrows(FireboltException.class, () -> connection.cancelAsyncQuery(token));
+            assertTrue(exception.getMessage().contains("Attempt to cancel query with query_id"));
```
goprean, line 187: does the message contain some wording that the query already finished so cannot cancel it? If so we should test it here since this message will be seen by the user.

goprean (approved):

src/integrationTest/java/integration/tests/AsyncQueryTest.java
bogdantruta-firebolt: This is the whole message form the server: "Attempt to cancel query with query_id = 5a21a8bb-9e2c-4ee6-81e1-2a875f5c3ffb failed. Query with such identifier does not exist. Use different query_id and re-run the statement." There is no way to know way to know if the query actually finished, unless we check the status when we search for the query_id before hand. I think that could be a nice improvement actually.

src/main/java/com/firebolt/jdbc/connection/FireboltConnection.java
```diff
@@ -686,4 +687,54 @@ public void register(CacheListener listener) {
...
+					throw new FireboltException("Could not get query_id for the async query with token: " + asyncQueryToken);
+				}
+				asyncQueryId = rs.getString("query_id");
+			}
+		}
+		if (StringUtils.isNotBlank(asyncQueryId)) {
+			try (PreparedStatement statement = createPreparedStatement("CANCEL QUERY WHERE query_id = ?")) {
+				statement.setString(1, asyncQueryId);
+				statement.executeUpdate();
+			}
```
ptiurin, line 717: Do we silently fail here? Let's try to raise the error so the user knows cancelling a query did not succeed.

bogdantruta-firebolt: We actually don't silently fail. If the cancel query fails, the error is not being caught anywhere, so it will propagate. there is a test for it in AsyncQueryTest

ptiurin (approved):
lgtm

PR #530, opened 2025-03-27
feat: FIR-44260 add connection info to user agent header

goprean:
Adde connection information to the user-agent hearer:
- if connection is not cached and created the first time then only: connId:<xxx> will be added
- if the connection was using cached data then we add: connId:<abc>;cachedConnId:<yyy> where yyy is the connection id that tested the connection. 

Seems like integration tests are all using cached connections (which is a good thing) and they are passing:

### DIFF
--- a/src/integrationTest/java/integration/tests/client/UsageTrackingTest.java
+++ b/src/integrationTest/java/integration/tests/client/UsageTrackingTest.java
@@ -10,6 +10,8 @@ import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
@@ -20,6 +22,7 @@ public class UsageTrackingTest extends MockWebServerAwareIntegrationTest {
 		mockBackEnd.enqueue(new MockResponse().setResponseCode(200));
 		mockBackEnd.enqueue(new MockResponse().setResponseCode(200));
 
+		String firstConnectionId;
 		try (FireboltConnection fireboltConnection = (FireboltConnection) createLocalConnection(String.format(
 				"?ssl=0&port=%d&max_retries=%d&user_drivers=AwesomeDriver:1.0.1&user_clients=GreatClient:0.1.4&access_token=token1",
 				mockBackEnd.getPort(), 3)); Statement statement = fireboltConnection.createStatement()) {
@@ -33,6 +36,9 @@ public class UsageTrackingTest extends MockWebServerAwareIntegrationTest {
 			// connection is not cached so there should be only connId set
 			assertTrue(userAgentHeader.contains("connId"));
 			assertFalse(userAgentHeader.contains("cachedConnId"));
+
+			firstConnectionId = userAgentHeader.split("connId:")[1];
+			assertEquals(12, firstConnectionId.length());
 		}
 
 		// create a second connection
@@ -48,7 +54,18 @@ public class UsageTrackingTest extends MockWebServerAwareIntegrationTest {
 
 			// connection should be cached
 			assertTrue(userAgentHeader.contains("connId"));
-			assertTrue(userAgentHeader.contains("cachedConnId"));
+			assertTrue(userAgentHeader.contains("cachedConnId:" + firstConnectionId));
+
+			// get the second connection id
+			// split the string based on cacheConnId and get the first part
+			String userAgentHeaderWithoutCachedConnectionId = userAgentHeader.split(";cachedConnId")[0];
+
+			// the split it based on connId: and get the second part
+			String secondConnectionId = userAgentHeaderWithoutCachedConnectionId.split("connId:")[1];
+			assertEquals(12, secondConnectionId.length());
+
+			// each connection should get their own ids
+			assertNotEquals(firstConnectionId, secondConnectionId);
 		}
 
 	}
--- a/src/integrationTest/java/integration/tests/client/UsageTrackingTest.java
+++ b/src/integrationTest/java/integration/tests/client/UsageTrackingTest.java
@@ -9,6 +9,7 @@ import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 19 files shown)


goprean:
Screenshot 2025-03-27 at 15 08 39
Screenshot 2025-03-27 at 15 02 39

src/test/java/com/firebolt/jdbc/connection/FireboltConnectionServiceSecretTest.java
```diff
@@ -405,12 +409,50 @@ void willGetJwtSystemEngineUrlUserEngineEndpointAndDatabaseFromCacheWhenConnecti
...
+        }
+    }
+
+    @Test
+    void willAddUserAgentHeaderWhenConnectionIsCachedByAPreviousConnection() throws SQLException {
+        // connection is already cached
+        ConnectionCache connectionCache = new ConnectionCache(ANOTHER_CONNECTION_ID);
+        connectionCache.setCacheSource(CacheType.MEMORY.name().toLowerCase());
+
+        lenient().when(mockCacheService.get(cacheKey)).thenReturn(Optional.of(connectionCache));
```
bogdantruta-firebolt, line 437: here as well

src/test/java/com/firebolt/jdbc/connection/LocalhostFireboltConnectionTest.java
```diff
@@ -161,9 +176,48 @@ void canCacheConnectionsToLocalhost(String actualValue, boolean expectedValue) t
...
+        disableCacheConnection();
+        try (FireboltConnection fireboltConnection = createConnection(LOCAL_URL, connectionProperties)) {
+            assertTrue(fireboltConnection.getConnectionUserAgentHeader().isEmpty());
+        }
+    }
+
+    @Test
+    void willAddUserAgentHeaderWhenConnectionIsCachedByThisConnection() throws SQLException {
+        // no cache is present for the key
+        lenient().when(cacheService.get(cacheKey)).thenReturn(Optional.empty());
```
bogdantruta-firebolt, line 190: here too

src/test/java/com/firebolt/jdbc/connection/LocalhostFireboltConnectionTest.java
```diff
@@ -161,9 +176,48 @@ void canCacheConnectionsToLocalhost(String actualValue, boolean expectedValue) t
...
+        }
+    }
+
+    @Test
+    void willAddUserAgentHeaderWhenConnectionIsCachedByAPreviousConnection() throws SQLException {
+        // connection is already cached
+        ConnectionCache connectionCache = new ConnectionCache(CONNECTION_ID_2);
+        connectionCache.setCacheSource(CacheType.MEMORY.name().toLowerCase());
+
+        lenient().when(cacheService.get(cacheKey)).thenReturn(Optional.of(connectionCache));
```
bogdantruta-firebolt, line 204: here too

src/test/java/com/firebolt/jdbc/client/UsageTrackerUtilTest.java
ptiurin, line 1: Let's add here a user agent test where connection string is provided, for completeness.

src/main/java/com/firebolt/jdbc/connection/ConnectionIdGenerator.java
ptiurin, line 1: Why do we need a class for this? Can't each connection instance generate an id independently?

src/integrationTest/java/integration/tests/client/UsageTrackingTest.java
```diff
@@ -17,16 +18,39 @@ public class UsageTrackingTest extends MockWebServerAwareIntegrationTest {
...
 			RecordedRequest request = mockBackEnd.takeRequest();
 			String userAgentHeader = request.getHeaders().get("User-Agent");
 			assertNotNull(userAgentHeader);
 			assertTrue(userAgentHeader.startsWith("GreatClient/0.1.4" + " JDBC/" + VersionUtil.getDriverVersion()));
-			assertTrue(userAgentHeader.endsWith("AwesomeDriver/1.0.1"));
+			assertTrue(userAgentHeader.contains("AwesomeDriver/1.0.1"));
+
+			// connection should be cached
+			assertTrue(userAgentHeader.contains("connId"));
+			assertTrue(userAgentHeader.contains("cachedConnId"));
```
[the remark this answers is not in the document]
goprean: Done. I have added some checks for the connection ids that are different and that the cachedConnId is the same as the connection first connection id

src/main/java/com/firebolt/jdbc/connection/ConnectionIdGenerator.java
goprean: They will, but it is easier to test it when we can mock it. Also if we want to have different implementations for example one that would generate a UUID and one that would generate a 12 chars random string like I have would be easier to switch.

src/test/java/com/firebolt/jdbc/client/UsageTrackerUtilTest.java
goprean: UsageTrackerUtil does not pass in the connection string. That is already tested in the FireboltClient.

bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/connection/ConnectionIdGenerator.java
goprean: Test Driven Design will impose your class design in a lot of ways. I feel that if we can have tests clearer and simpler it is always worth the effort of designing the classes that way. There is a good blog pos from google that explains why it is never a good practice to create objects in code but to use dependency injection and let other objects know how to create the objects you need (I need to look it up)

src/main/java/com/firebolt/jdbc/cache/key/LocalhostCacheKey.java
ptiurin, line 1: Do we even need to cache anything for the localhost? I would imagine there's no latency when run locally.

ptiurin (approved):
Otherwise looks good

src/main/java/com/firebolt/jdbc/cache/key/LocalhostCacheKey.java
goprean: My thought here was that we can test the caching functionality (creating the file, deleting the file, etc) for free when the localhost testing is done through the team that is using the JDBC driver for local testing.

ptiurin: My only worry here would be the test scenarios they might be running could be affected by this feature. Let's check that this won't break anything on PackDB side.

goprean: I will check that this feature will not break anything on their side yes. I need to figure out first how to get the instructions to run their jobs using latest jdbc jar. That part I don't know how to do

PR #531, opened 2025-03-31
feat: FIR-44732 Support LocalDate and LocalDateTime in prepared statements

bogdantruta-firebolt:
Created a patch that allows users to set LocalDate and LocalDateTime object values into PreparedStatements.

### DIFF
--- a/src/integrationTest/java/integration/tests/PreparedStatementTest.java
+++ b/src/integrationTest/java/integration/tests/PreparedStatementTest.java
@@ -38,6 +38,8 @@ import java.sql.Timestamp;
 import java.sql.Types;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -430,7 +432,6 @@ class PreparedStatementTest extends IntegrationTest {
 	void shouldInsertAndSelectStruct() throws SQLException {
 		Car car1 = Car.builder().make("Ford").sales(12345).ts(new Timestamp(2)).d(new Date(3)).build();
 
-		executeStatementFromFile("/statements/prepared-statement/ddl.sql");
 		try (Connection connection = createConnection()) {
 
 			try (PreparedStatement statement = connection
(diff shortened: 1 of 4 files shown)


ptiurin (commented):
Let's add some integration tests too so we're not just testing against mocks. Otherwise LGTM.

src/integrationTest/java/integration/tests/PreparedStatementTest.java
```diff
@@ -457,6 +459,31 @@ void shouldInsertAndSelectStruct() throws SQLException {
...
 	}
 
+	@ParameterizedTest
+	@MethodSource("dateTypes")
+	void shouldFetchTimestampAndDate(Object timestampOrLocalDateTime, Object dateOrLocalDate, boolean addTargetSqlType) throws SQLException {
+		String expectedTimestamp = "2019-07-31 14:15:13";
+		String expectedDate = "2019-07-31";
+		try (Connection connection = createConnection()) {
+			try (PreparedStatement statement = connection
+					.prepareStatement("SELECT ? as a, ? as b")) {
```
goprean, line 469: Is this how we are testing the prepared statement using select? I am much more in favor of creating tables with certain columns type and use prepared statements to insert and to get the data from those columns.

src/integrationTest/java/integration/tests/PreparedStatementTest.java
```diff
@@ -555,6 +582,19 @@ void shouldFetchSpecialCharacters() throws SQLException, MalformedURLException {
 		}
 	}
 
+	Stream<Arguments> dateTypes() {
+		return Stream.of(
+				Arguments.of(LocalDateTime.of(2019, 7, 31, 14, 15, 13),
+						LocalDate.of(2019, 7, 31), true),
+				Arguments.of(new Timestamp(1564571713000L),
+						new Date(1564527600000L), true),
+				Arguments.of(LocalDateTime.of(2019, 7, 31, 14, 15, 13),
```
goprean, line 591: I see that you have added the OffsetDateTime. Let's add some tests for that conversion as well

bogdantruta-firebolt: I removed it since it didn't work properly and this implementation is out of the scope of this PR. Will do it properly in the future

PR #532, opened 2025-03-31
fix: FIR-44784 Fix boolean in prepared statements

bogdantruta-firebolt:
Added a specialized class to handle all sorts of conversions from String, Numbers and Characters to Boolean.
Created tests for this method but in the current configuration of our transformer they do not work. 
We will need to update it to align ourself with the JDBC spec.
For the moment, setObject for a boolean outcome only work from a boolean object.

### DIFF
--- a/src/integrationTest/java/integration/tests/PreparedStatementTest.java
+++ b/src/integrationTest/java/integration/tests/PreparedStatementTest.java
@@ -12,9 +12,9 @@ import lombok.Builder;
 import lombok.CustomLog;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.junit.Ignore;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -612,9 +612,9 @@ class PreparedStatementTest extends IntegrationTest {
 		}
 	}
 
-	@Ignore
+	@Disabled
 	@ParameterizedTest
-	@MethodSource("booleanTypes")
+	@MethodSource("com.firebolt.jdbc.testutils.TestFixtures#booleanTypes")
 	void shouldFetchBooleanFromVariousObjects(Object objectTrue, Object objectFalse) throws SQLException {
 		try (Connection connection = createConnection()) {
 			try (PreparedStatement statement = connection
@@ -632,26 +632,6 @@ class PreparedStatementTest extends IntegrationTest {
 				assertFalse(rs.getBoolean(2));
 			}
 		}
-	}
-
-	Stream<Arguments> booleanTypes() {
-		return Stream.of(
-				Arguments.of("1", "0"),
-				Arguments.of("true", "false"),
-				Arguments.of("t", "f"),
-				Arguments.of("yes", "no"),
-				Arguments.of("y", "n"),
-				Arguments.of("on", "off"),
-				Arguments.of('1', '0'),
-				Arguments.of('t', 'f'),
-				Arguments.of('T', 'F'),
-				Arguments.of('y', 'n'),
-				Arguments.of('Y', 'N'),
-				Arguments.of(1, 0),
-				Arguments.of(1.0, 0.0),
-				Arguments.of(1F, 0F),
-				Arguments.of(1L, 0L)
-		);
 	}
 
 	Stream<Arguments> dateTypes() {
--- a/src/integrationTest/java/integration/tests/PreparedStatementTest.java
+++ b/src/integrationTest/java/integration/tests/PreparedStatementTest.java
@@ -12,6 +12,7 @@ import lombok.Builder;
 import lombok.CustomLog;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.junit.Ignore;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
@@ -38,8 +39,6 @@ import java.sql.Timestamp;
 import java.sql.Types;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
-import java.time.OffsetDateTime;
-import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


src/integrationTest/java/integration/tests/PreparedStatementTest.java
```diff
@@ -555,6 +556,72 @@ void shouldFetchSpecialCharacters() throws SQLException, MalformedURLException {
 		}
 	}
 
+	@Test
+	void shouldFetchBoolean() throws SQLException {
+		try (Connection connection = createConnection()) {
+			try (PreparedStatement statement = connection
+					.prepareStatement("SELECT ? as a, ? as b, ? as c")) {
+				statement.setBoolean(1, true);
+				statement.setObject(2, false);
```
goprean, line 565: the bug mentions that this does not work: statement.setObject(paramIndex, 1,  SqlTypes.BOOLEAN). 
Can we also set it in tests to make sure it works?

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param in Object to cast into boolean
+     * @return boolean value corresponding to the cast of the object
+     * @throws FireboltException cannot
+     */
+    @SuppressWarnings("java:S6201")
+    static boolean castToBoolean(final Object in) throws FireboltException {
+        log.debug("Cast to boolean: \"{0}\"", String.valueOf(in));
+        if (in instanceof Boolean) {
+            return (Boolean) in;
```
goprean, line 29: I think we need to check for nulls. If you convert a null to a boolean it will throw a NPE (null pointer exception)

src/test/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatementTest.java
```diff
@@ -403,7 +403,7 @@ void shouldSetBoolean() throws SQLException {
 		statement.execute();
 
 		verify(fireboltStatementService).execute(queryInfoWrapperArgumentCaptor.capture(), eq(properties), any());
-		assertEquals("INSERT INTO cars(available) VALUES (1)", queryInfoWrapperArgumentCaptor.getValue().getSql());
```
goprean, line 406: why are we changing this? the issue was with boolean as a number of (in this case 1). We should probably add another insert with VALUES(true) but as a separate line.

src/test/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatementTest.java
```diff
@@ -542,7 +542,7 @@ void shouldSetAllObjects() throws SQLException {
 
 		verify(fireboltStatementService).execute(queryInfoWrapperArgumentCaptor.capture(), eq(properties), any());
 		assertEquals(
-				"INSERT INTO cars(timestamp, date, float, long, big_decimal, null, boolean, int) VALUES ('2019-07-31 12:15:13','2019-07-31',5.5,5,555555555555.55555555,NULL,1,5)",
+				"INSERT INTO cars(timestamp, date, float, long, big_decimal, null, boolean, int) VALUES ('2019-07-31 12:15:13','2019-07-31',5.5,5,555555555555.55555555,NULL,true,5)",
```
goprean, line 545: same here. Why are we changing it from 1 to true. The problem was with 1.

src/test/java/com/firebolt/jdbc/statement/preparedstatement/FireboltPreparedStatementTest.java
```diff
@@ -567,7 +567,7 @@ void shouldSetAllObjectsWithCorrectSqlType() throws SQLException {
 		verify(fireboltStatementService).execute(queryInfoWrapperArgumentCaptor.capture(), eq(properties), any());
 
 		assertEquals(
-				"INSERT INTO cars(timestamp, date, float, long, big_decimal, null, boolean, int) VALUES ('2019-07-31 12:15:13','2019-07-31',5.5,5,555555555555.55555555,NULL,1,5)",
+				"INSERT INTO cars(timestamp, date, float, long, big_decimal, null, boolean, int) VALUES ('2019-07-31 12:15:13','2019-07-31',5.5,5,555555555555.55555555,NULL,true,5)",
```
goprean, line 570: and here

goprean (commented):
I see that in some tests you have changed from 1 to true (boolean). But wasn't the actual problem that 1 was not treated as a boolean?

src/integrationTest/java/integration/tests/PreparedStatementTest.java
bogdantruta-firebolt: I test that in the following 2 lines, both the paramIndex, object method, and the one with the sqlTargetType, but I could make a test where all values are both true and false to test both cases for all methods

src/integrationTest/java/integration/tests/PreparedStatementTest.java
```diff
@@ -555,6 +556,72 @@ void shouldFetchSpecialCharacters() throws SQLException, MalformedURLException {
...
+				assertEquals(FireboltDataType.BOOLEAN.name().toLowerCase(),
+						rs.getMetaData().getColumnTypeName(2).toLowerCase());
+				assertTrue(rs.next());
+				assertTrue(rs.getBoolean(1));
+				assertFalse(rs.getBoolean(2));
+			}
+		}
+	}
+
+	Stream<Arguments> booleanTypes() {
```
ptiurin, line 605: We're duplicating this method. Is there a way to reuse it in different tests? Can we add it to the parent class?

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Cast an Object value to the corresponding boolean value.
+     *
+     * @param in Object to cast into boolean
+     * @return boolean value corresponding to the cast of the object
+     * @throws FireboltException cannot
+     */
+    @SuppressWarnings("java:S6201")
```
ptiurin, line 25: What are we suppressing here?

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if (in instanceof Number) {
+            return fromNumber((Number) in);
+        }
+        throw new FireboltException(format("Cannot cast %s to boolean", in));
+    }
+
+    static boolean fromString(final String strval) throws FireboltException {
+        // Leading or trailing whitespace is ignored, and case does not matter.
+        final String val = strval.trim();
```
ptiurin, line 45: I'm not sure if we should do this. If there's whitespace something clearly is wrong and we should not assume the value is boolean. Instead, make the user sanitize the values.

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
```diff
@@ -0,0 +1,88 @@
...
+        if ("1".equals(val) || "true".equalsIgnoreCase(val)
+                || "t".equalsIgnoreCase(val) || "yes".equalsIgnoreCase(val)
+                || "y".equalsIgnoreCase(val) || "on".equalsIgnoreCase(val)) {
+            return true;
+        }
+        if ("0".equals(val) || "false".equalsIgnoreCase(val)
+                || "f".equalsIgnoreCase(val) || "no".equalsIgnoreCase(val)
+                || "n".equalsIgnoreCase(val) || "off".equalsIgnoreCase(val)) {
```
ptiurin, line 53: I'm not on board with allowing on/off and y/n. Is there a convention on which values are castable to boolean?

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
bogdantruta-firebolt: I believe that since we are casting other objects to a boolean value, we might as well trim whitespaces from a string. This behavior is taken directly from the Postgres JDBC driver

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
bogdantruta-firebolt: I am not sure of any convention, but after trying these values directly in the Firebolt console, they work in the same way as inputing the related boolean value, so I am guessing changing them client side would not do us any harm. Then again, we have the option of just setting the number/text value, but that would not work in the case of a select statement,

src/main/java/com/firebolt/jdbc/type/BooleanTypeUtil.java
ptiurin: Ok, looks like PG also thinks this is fine so I guess we'll do the same.

ptiurin (approved):
LGTM, but make sure to resolve any open conversations.

goprean (approved):
I think the spec is very clear about how the transforming should be done. 
If setObject(index, object) is used without the sql type, then a default mapping is done (table B4 from the spec: https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf?AuthParam=1744031434_3e1d01e62d033e97eb5e7c88f5039beb). For example if the object is Boolean.class then we transform it to a boolean value to be sent to firebolt. 

if setObject(index, object, sqlType) is used, then the spec is also clear that only certain java classes with sql types should match. (the full list is B5 table - https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf?AuthParam=1744031434_3e1d01e62d033e97eb5e7c88f5039beb).

So while this is fixing this case, I can see other cases where the setObject method is not going to work.

PR #536, opened 2025-04-04
FID-44260 added class to resolve default temp directory

goprean:
Added the class that will resolve the default temp directory for writing cache files. 

We have to change some environment variables so I had to change the test setup. Since our tests are run in parallel, and we change the default temp directory some tests will fail since Mockito will try to run to the same temp directory. So when changing this class we need to run manually the tests in the DirectoryPathResolverTest (notice how the tests are disabled so they won't run In pipeline)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ test {
             "--add-opens=java.base/java.util=ALL-UNNAMED"
     ) // so we can change the environment variables
     useJUnitPlatform()
-    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    maxParallelForks = 1
     testLogging {
         events 'PASSED', 'FAILED', 'SKIPPED'
         exceptionFormat = 'full'
--- a/build.gradle
+++ b/build.gradle
@@ -102,6 +102,10 @@ dependencies {
 }
 
 test {
+    jvmArgs(
+            "--add-opens=java.base/java.lang=ALL-UNNAMED",
+            "--add-opens=java.base/java.util=ALL-UNNAMED"
+    ) // so we can change the environment variables
     useJUnitPlatform()
     maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
     testLogging {
@@ -224,6 +228,7 @@ task printVersion {
 }
 sonarqube {
     properties {
+        property "sonar.java.source", "11"
         property "sonar.projectKey", "firebolt-db_jdbc"
         property "sonar.organization", "firebolt-db"
         property "sonar.host.url", "https://sonarcloud.io"
--- a/src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
+++ b/src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
@@ -1,0 +1,63 @@
+package com.firebolt.jdbc.cache;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.SetEnvironmentVariable;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+// The tests are disabled since they interfere with the other test classes (tests are run in parallel and mockito creates file in the temp location,
+// which we modify in these tests, so the other tests would randomly fail)
+@Disabled
+class DirectoryPathResolverTest {
+
+    private static String originalUserHome;
+    private static String originalOsName;
+
+    private DirectoryPathResolver directoryPathResolver = new DirectoryPathResolver();
+
+    @BeforeAll
+    static void setupClass() {
+        originalUserHome = System.getProperty(DirectoryPathResolver.USER_HOME_PROPERTY);
+        originalOsName = System.getProperty(DirectoryPathResolver.OS_NAME_PROPERTY);
+    }
+
+    @AfterAll
+    static void tearDownClass() {
+        // set back the original properties
+        System.setProperty(DirectoryPathResolver.USER_HOME_PROPERTY, originalUserHome);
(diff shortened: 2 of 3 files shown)


bogdantruta-firebolt (approved):
lgtm

src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
```diff
@@ -0,0 +1,63 @@
...
+        System.setProperty(DirectoryPathResolver.OS_NAME_PROPERTY,"windows 11");
+        System.setProperty(DirectoryPathResolver.TEMP_DIRECTORY_PROPERTY, "/temp");
+        assertEquals("/temp/fireboltDriverJdbc", directoryPathResolver.resolveFireboltJdbcDirectory().toString());
+    }
+
+    @Test
+    @SetEnvironmentVariable(key="TMPDIR", value="/temp/xx/1234")
+    void canCreateFireboltDirectoryForMac() {
+       System.setProperty(DirectoryPathResolver.OS_NAME_PROPERTY,"mac osx 18");
+       assertEquals("/temp/xx/1234/fireboltDriverJdbc", directoryPathResolver.resolveFireboltJdbcDirectory().toString());
```
bogdantruta-firebolt, line 45: the path is no longer fireboltDriverJdbc, but fireboltDriver

bogdantruta-firebolt (changes requested):

src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.SetEnvironmentVariable;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+// The tests are disabled since they interfere with the other test classes (tests are run in parallel and mockito creates file in the temp location,
+// which we modify in these tests, so the other tests would randomly fail)
+@Disabled
+class DirectoryPathResolverTest {
```
ptiurin, line 14: > So when changing this class we need to run manually the tests in the DirectoryPathResolverTest (notice how the tests are disabled so they won't run In pipeline)

Can we try to find a way to execute these tests sequentially? I'm sure we'll quickly forget about the special case here that requires manual test run and rely on automated testing to verify everything works.
As a last resort we can probably modify GH test workflow to run this test separately from the rest.

build.gradle
```diff
@@ -223,6 +227,7 @@ task printVersion {
 }
 sonarqube {
     properties {
+        property "sonar.java.source", "11"
```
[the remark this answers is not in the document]
goprean: Yes. It tells sonar to use the rules for Java 11. In my case it was complaining that instead of ..Stream.collect(Collectors.toList()) to use Stream.toList(), but that is only available in Java 16.

src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
goprean: I have added a nightly job to test the matrix combinations: os & java version. It will call the existing integration tests.

goprean: Also enabled back the DirectoryPathResoverTest and made the test run sequentially.

src/test/java/com/firebolt/jdbc/cache/DirectoryPathResolverTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.SetEnvironmentVariable;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+// The tests are disabled since they interfere with the other test classes (tests are run in parallel and mockito creates file in the temp location,
+// which we modify in these tests, so the other tests would randomly fail)
```
ptiurin, line 11: Let's remove/modify this comment since the tests are no longer disabled. This way we don't get confused later if they supposed to run or not.

goprean: Done.

.github/workflows/nightly.yaml
ptiurin, line 1: As discussed, let's split it into v1 and v2 counterparts.

goprean: Done. Split it into v1 and v2. Can we test this while not in master? I don't see the GitHub action for nightly-v1 and nightly-v2

goprean: I have changed the integration action as well so that won't work. But to me it should be fine to push it to master and refine it afterwards. Would that work for you?

goprean: Or another option would be to create separate downstream jobs as well and keep the ones we have intact (for integration tests and for v1 and v2). That way we don't interrupt the current flow (anyhow the integration tests are run manually so we are not really interrupting anything). and test the new ones until we know that they do what they  have to do.

goprean: Even changing the performance tests will not work as it will try to make commit the changes in a new PR. I would suggest we move with this change and I can test it and fix it in master.

bogdantruta-firebolt (approved):

.github/workflows/nightly.yaml
ptiurin: > I have changed the integration action as well so that won't work. 
Integration action will be run from the branch that you specify, it uses the context of where you've submitted it.

ptiurin: Let's sync and I can show you.

ptiurin:
@goprean test

.github/workflows/nightly.yaml
goprean: I have removed the nightly since they were committed as part of its own PR.

PR #537, opened 2025-04-04
FIR-44260 added encryption service

goprean:
Added the class to encrypt and decrypt using AES/GCM algorithm. 
The encryption will be used in 2 places: to encrypt the file name and to encrypt the content of the file
The decryption will be used to decrypt the content of the file.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
+++ b/src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
@@ -44,7 +44,7 @@ public class EncryptionService {
             return encryptAESGCM(plainText, encryptionKey);
         } catch (Exception e) {
             log.error("Failed to encrypt the text", e);
-            throw EncryptionException.encryptionFailed();
+            throw new EncryptionException("Encryption failed.");
         }
     }
 
@@ -64,7 +64,7 @@ public class EncryptionService {
             return decryptAESGCM(base64EncryptedText, encryptionKey);
         } catch (Exception e) {
             log.error("Failed to decrypt encrypted text", e);
-            throw EncryptionException.decryptionFailed();
+            throw new EncryptionException("Decryption failed.");
         }
     }
 
--- a/src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
+++ b/src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
@@ -1,0 +1,138 @@
+package com.firebolt.jdbc.cache;
+
+import com.firebolt.jdbc.cache.exception.EncryptionException;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import lombok.CustomLog;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Class that knows how to encrypt and encrypt secrets using AES encryption algorithm
+ */
+@CustomLog
+public class EncryptionService {
+
+    private static final int GCM_TAG_LENGTH = 128; // GCM tag length in bits (16 bytes)
+    private static final int AES_KEY_SIZE = 32; // use 32 bytes so we will use AES-256 algorithm
+    private static final String SHA_256_ALGO = "SHA-256";
+
+    /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+@CustomLog
+public class EncryptionService {
+
+    private static final int GCM_TAG_LENGTH = 128; // GCM tag length in bits (16 bytes)
+    private static final int AES_KEY_SIZE = 32; // use 32 bytes so we will use AES-256 algorithm
+    private static final String SHA_256_ALGO = "SHA-256";
+
+    /**
+     * Encrypts a plaintext using the encryption key. If there are any exceptions during encryption an empty optional will be returned
```
bogdantruta-firebolt, line 31: In the case of exceptions during encryption we throw a the exception further from what I see

src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Decrypts the encrypted text using AES algorithm
+     * @param base64EncryptedText - the base64 encoded string
+     * @param encryptionKey - the key to decrypt
+     * @return the original plain text
+     * @throws Exception - in case the encrypted text cannot be decrypted
+     */
+    public String decrypt(String base64EncryptedText, String encryptionKey) throws EncryptionException, IllegalArgumentException {
+        if (StringUtils.isBlank(base64EncryptedText) || StringUtils.isBlank(encryptionKey)) {
+            throw new IllegalArgumentException("Text to decrypt or encryption key is null. Cannot encrypt.");
```
bogdantruta-firebolt, line 59: exception should state cannot decrypt

src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private Cipher getAesGcmCipherInstance() throws NoSuchPaddingException, NoSuchAlgorithmException {
+        return Cipher.getInstance("AES/GCM/NoPadding");
+    }
+
+    // Deterministic nonce generation using SHA-256 hash
+    private static byte[] deriveNonce(String plaintext, String key) throws NoSuchAlgorithmException {
+        MessageDigest digest = MessageDigest.getInstance(SHA_256_ALGO);
+        byte[] hash = digest.digest((plaintext + key).getBytes(StandardCharsets.UTF_8));
```
bogdantruta-firebolt, line 133: do we still need the two params if nonce is derived from just the key? We could repeat the key value or if possible just use it once

bogdantruta-firebolt (approved):
lgtm other than comms

src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
goprean: Fixed

src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
goprean: Fixed

src/main/java/com/firebolt/jdbc/cache/EncryptionService.java
goprean: Done.

src/test/java/com/firebolt/jdbc/cache/EncryptionServiceTest.java
ptiurin, line 1: Let's also add a check that encrypted value != unencrypted value. Otherwise right now if encrypt and decrypt return the same initial string these tests will pass.

src/main/java/com/firebolt/jdbc/cache/exception/EncryptionException.java
```diff
@@ -0,0 +1,12 @@
...
+public class EncryptionException extends RuntimeException {
+
+    public static EncryptionException encryptionFailed() {
+        return new EncryptionException();
+    }
+
+    public static EncryptionException decryptionFailed() {
+        return new EncryptionException();
+    }
+}
```
ptiurin, line 12: What do we get from having these methods? Any advantage over having `EncryptionException` and `DecryptionException` or `EncryptionException(message)`?

goprean: Initially I thought this will be getting to the enduser and I thought if we ever need i18n (internationalization of the message) the message key would be resolved here. But these messages will not get to the enduser (we are logging the message). I will consolidate it to just one constructor with a string.

goprean: Done

src/test/java/com/firebolt/jdbc/cache/EncryptionServiceTest.java
goprean: Done

ptiurin (approved):
lgtm

PR #538, opened 2025-04-04
FIR-44260: added the file service and filename generator service

goprean:
Added the FileService (knows how to interact with the files on disk: read, write, delete, etc)
Added the FIlenameGenerator that will take a cacheKey and generate the file name for it. 
Also added the encryption key for each of the CacheKey, so the CacheKey is responsible to provide the encryption key.

### DIFF
--- a/src/main/java/com/firebolt/jdbc/cache/FileService.java
+++ b/src/main/java/com/firebolt/jdbc/cache/FileService.java
@@ -1,0 +1,159 @@
+package com.firebolt.jdbc.cache;
+
+import com.firebolt.jdbc.annotation.ExcludeFromJacocoGeneratedReport;
+import com.firebolt.jdbc.cache.exception.ConnectionCacheDeserializationException;
+import com.firebolt.jdbc.cache.exception.EncryptionException;
+import com.firebolt.jdbc.cache.exception.FilenameGenerationException;
+import com.firebolt.jdbc.cache.key.CacheKey;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.FileTime;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import lombok.CustomLog;
+import org.json.JSONObject;
+
+@CustomLog
+public class FileService {
+
+    private DirectoryPathResolver directoryPathResolver;
+    private FilenameGenerator filenameGenerator;
+    private ExecutorService executorService;
+    private EncryptionService encryptionService;
+
+    private static FileService instance;
+
+    private Path fireboltJdbcDirectory;
+
+    @ExcludeFromJacocoGeneratedReport
+    public static FileService getInstance() {
+        if (instance == null) {
+            instance = new FileService(new DirectoryPathResolver(), new FilenameGenerator(), new EncryptionService(), Executors.newFixedThreadPool(2));
+        }
+        return instance;
+    }
+
+    // visible for testing
+    FileService(DirectoryPathResolver directoryPathResolver, FilenameGenerator filenameGenerator, EncryptionService encryptionService, ExecutorService executorService) {
+        this.directoryPathResolver = directoryPathResolver;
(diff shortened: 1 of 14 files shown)


src/test/java/com/firebolt/jdbc/cache/FileServiceTest.java
```diff
@@ -0,0 +1,253 @@
...
+            filesMockedStatic.when(() -> Files.writeString(any(Path.class), anyString(), eq(StandardOpenOption.TRUNCATE_EXISTING))).thenThrow(IOException.class);
+
+            fileService.safelyWriteFile(mockFile, ENCRYPTED_CONTENT);
+
+            filesMockedStatic.verify(() -> Files.getAttribute(mockPath, "basic:creationTime"));
+        }
+    }
+
+    @Test
+    void willWriteToDisk() {
```
bogdantruta-firebolt, line 210: I think this does not fully write to disk since it throws exception when writing to file

src/test/java/com/firebolt/jdbc/cache/FileServiceTest.java
```diff
@@ -0,0 +1,253 @@
...
+        try (MockedStatic<Files> filesMockedStatic = mockStatic(Files.class)) {
+            Path filePathThatDoesNotExist = Path.of("does", "not", "exist");
+            filesMockedStatic.when(() -> Files.deleteIfExists(filePathThatDoesNotExist)).thenReturn(false);
+            fileService.safelyDeleteFile(filePathThatDoesNotExist);
+            filesMockedStatic.verify(() -> Files.deleteIfExists(filePathThatDoesNotExist));
+        }
+    }
+
+    @SuppressWarnings("java:S2925")
+    private void sleepForMillis(long millis) {
```
bogdantruta-firebolt, line 235: Could we maybe introduce a TestHelper/ TestFixtures class in testCommons which could have common methods/ argument providers like this one?

src/test/java/com/firebolt/jdbc/cache/FileServiceTest.java
goprean: Nice catch. Thanks. I have added few more assertions and verifications.

src/test/java/com/firebolt/jdbc/cache/FileServiceTest.java
goprean: Usually we can have this type of things in a base class. that all the other tests can extend.

src/main/java/com/firebolt/jdbc/cache/FileService.java
```diff
@@ -0,0 +1,159 @@
...
+                    boolean fileCreated = file.createNewFile();
+                    if (!fileCreated) {
+                        log.error("Cannot create file to save the connection cache.");
+                        return;
+                    }
+                } catch (IOException e) {
+                    // maybe do not have permission to write to that location
+                    log.error("Cannot create on-disk connection cache. Maybe do not have the write permission. ", e);
+                    return;
+                }
```
ptiurin, line 91: Small point, but I wouldn't log them as errors here (and in other places in this class), since this is not a requirement for the user to have a writable FS and can cause confusion. "warning" at most.

ptiurin (commented):
apart from that - no issues

build.gradle
```diff
@@ -107,7 +107,7 @@ test {
             "--add-opens=java.base/java.util=ALL-UNNAMED"
     ) // so we can change the environment variables
     useJUnitPlatform()
-    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    maxParallelForks = 1
```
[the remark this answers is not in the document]
goprean: I'll revert. I started with the change from the other PR but I thought I reverted the gradle file. Weird. It was not meant to be in this review. Thanks for catching.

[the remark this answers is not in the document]
goprean: It now shows the changes from the previous files as well as I need to rebase the previous changes and only apply the current changes.

src/main/java/com/firebolt/jdbc/cache/FileService.java
goprean: Good point. Moved all the log.error to log.warn in the FileService class.

ptiurin (approved):

PR #540, opened 2025-04-07
test: Fix struct set commands in integration tests

ptiurin:
Fixing set commands that enable struct.


goprean (approved):
Thanks for fixing this Petro. The integration tests were failing due to this.

bogdantruta-firebolt (approved):

PR #542, opened 2025-04-09
test: Fixed the tests to compile with Java 11

goprean:
Background:
I was trying to add a nightly job to run the tests on 11, 17 and 21 (all the long term supported java versions) and I had realized that Java 11 the tests were not compiling. 
The """ multi string line concept was added in Java 15. So had to sent this extra review to fix the tests to run on Java 11 (since that is how we are testing the driver that it works on Java 11 by running the integration tests"


ptiurin (approved):
Lgtm, good to see the nightly already spotting issues!

PR #543, opened 2025-04-09
ci: FIR-45195 add nighlty job to run v1 and v2 integration tests

goprean:
I have added the nightly jobs as part of its own review (I will take it out of the PR that dealt with the default temp directory).

I have added just one combination (java 11 and ubuntu). When running multiple combinations I see that the tests are randomly failing. I need to investigate why, but we will still get some benefits of running v1 and v2 nightly on java 11 on ubuntu meanwhile. 

Few notes:
- we don't have a runner for Mac (at least when I had the Mac in the os matrix, it did not seem to run)
- I had to add shell: bash to the scripts that were doing some checking since those were failing on windows. 

Next steps. I want to have it run on Java 17 and 21 and also on windows-latest. I need to check why they are failing though. 

In order to test it I had to modify the perf git hub action and use it on my branch (See the screenshot below for v1 and v2 results of nightly runs)

### DIFF
--- a/.github/workflows/nightly-v2.yaml
+++ b/.github/workflows/nightly-v2.yaml
@@ -1,0 +1,23 @@
+name: Nightly run for integration tests against v2 firebolt
+
+on:
+  schedule:
+    - cron: '30 3 * * *'        # Daily at 03:30 UTC
+  workflow_dispatch:             # Allow manual trigger
+
+jobs:
+  nightly-integration-runs-v2:
+    strategy:
+      fail-fast: false # finish all jobs even if one fails
+      max-parallel: 2
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        java: [11, 17, 21]
+
+    uses: ./.github/workflows/integration-test-v2.yml
+    with:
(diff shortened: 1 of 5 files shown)


goprean:
Screenshot 2025-04-09 at 16 26 50
Screenshot 2025-04-09 at 16 22 38

.github/workflows/nightly-v2.yaml
ptiurin, line 1: I'd suggest adding
```
    strategy:
      fail-fast: false # finish all jobs even if one fails
      max-parallel: 2
```
max-parallel ensures we don't hammer the services too much, especially auth. Also since there's a limit on how many engines we can have running on an account at the same time this ensures we're not hitting that. For nightly we don't care about it being fast and can let it run so we check it out in the morning.
fail-fast: false ensures we run through all of them even if a pair fails.

goprean: Done. It's in progress and I can see only 2 jobs are running at a time: https://github.com/firebolt-db/jdbc/actions/runs/14361241038

ptiurin (approved):
lgtm

PR #544, opened 2025-04-09
test: fix the integration tests to run in parallel

goprean:
Problem: when trying to setup nightly with different os name and java versions only one run was passing. 

When we have 3 os (windows, ubuntu, and macOS) and 3 java versions: 11, 17 and 21 the nightly job was creating 9 jobs to run integration tests. There jobs would run in parallel. 

Engine name and db names are unique for each run, but some tests were creating the same engine name in the ddl. 

Since these tests ran in parallel, the first run that succeeded would delete the engines (but at that point the ddl might have already run for the other runs so the test was assuming that the engine is present), until it was deleted. 

Create the engines in tests since we can get the engine name from the system property and append some values to it. This would guarantee that the engine names are unique across parallel runs.

### DIFF
--- a/src/integrationTest/java/integration/tests/AsyncQueryTest.java
+++ b/src/integrationTest/java/integration/tests/AsyncQueryTest.java
@@ -4,21 +4,19 @@ import com.firebolt.jdbc.connection.FireboltConnection;
 import com.firebolt.jdbc.exception.FireboltException;
 import com.firebolt.jdbc.statement.FireboltStatement;
 import integration.IntegrationTest;
-import lombok.CustomLog;
-import org.apache.commons.lang3.StringUtils;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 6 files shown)


src/integrationTest/java/integration/tests/AsyncQueryTest.java
```diff
@@ -45,8 +43,14 @@ void afterEach() {
 
     @BeforeAll
     void beforeAll() {
+        engineName = System.getProperty("engine") + "async_test_second_engine";
```
bogdantruta-firebolt, line 46: I believe we should add the table here as well, since in the afterEach method, we delete all content from the table and this could interfere with other tests

goprean: The table is created inside a database. Now the databases are unique. so they should not interfere with other runs. If it runs fine in one run, it should be fine in parallel runs. Am I missing something?

bogdantruta-firebolt: You are right, I forgot about this :))

PR #546, opened 2025-04-10
(test) FIR-45231 - Fixed the TLS test that fails on windows

goprean:
On windows the TLS tests are failing with the following error
 javax.net.ssl.SSLPeerUnverifiedException: Hostname localhost not verified:
            certificate: sha256/yOowwRGffNWswtSf8MidkjubXGpM7kiG/amLJeRrOsw=
            DN: CN=5e8635da-3d3d-4f80-8293-1e70661934d0
            subjectAltNames: [127.0.0.1]

On Windows the resolution from localhost to 127.0.0.1 does not happen. So need to add both localhost and 127.0.0.1 as the subject alternative names for the cert.

### DIFF
--- a/src/integrationTest/java/integration/tests/client/TLSTest.java
+++ b/src/integrationTest/java/integration/tests/client/TLSTest.java
@@ -6,7 +6,6 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Writer;
-import java.net.InetAddress;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


bogdantruta-firebolt (approved):

src/integrationTest/java/integration/tests/client/TLSTest.java
```diff
@@ -25,19 +24,18 @@ public void shouldUseTLS() throws SQLException, IOException, NoSuchFieldExceptio
...
+				.addSubjectAlternativeName("localhost")
 				.build();
 
 		// HandshakeCertificates contains certificates for a TLS handshake
 		HandshakeCertificates serverCertificates = new HandshakeCertificates.Builder()
 				.heldCertificate(localhostCertificate).addPlatformTrustedCertificates().build();
 
 		mockBackEnd.useHttps(serverCertificates.sslSocketFactory(), false);
-
+
```
ptiurin, line 38: Let's try to avoid adding whitespace

ptiurin (approved):
Otherwise lgtm

src/integrationTest/java/integration/tests/client/TLSTest.java
goprean: Done

PR #554, opened 2025-04-14
test FIR-45277 fix the gradle script for windows v1 integration tests

goprean:
I have seen v1 scripts fail with the following error when running on Windows:

Task '.staging.firebolt.io' not found in root project 'firebolt-jdbc'.

On windows we need to pass in the env variables to Gradle in quotes. 

Windows run: https://github.com/firebolt-db/jdbc/actions/runs/14439806639 (it fails with  the timeout tests)
Ubuntu: https://github.com/firebolt-db/jdbc/actions/runs/14439813744
MacOs: https://github.com/firebolt-db/jdbc/actions/runs/14439817374/job/40487329231

### DIFF
--- a/.github/workflows/integration-test-v1.yml
+++ b/.github/workflows/integration-test-v1.yml
@@ -104,4 +104,4 @@ jobs:
         run: chmod +x gradlew
 
       - name: Run integration tests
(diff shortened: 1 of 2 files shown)


ptiurin (approved):

ptiurin:
Should we do the same for v2?

goprean:
> Should we do the same for v2?

v2 was not passing any unquoted parameters that would cause issue. But to keep it consistent I have quoted all the params for v2 as well

PR #555, opened 2025-04-14
(test) FIR-45277 pass in max series size from GitHub

goprean:
The last test that is failing on nightly (both v1 and v2) is the timeout test. 

There are 2 types of failures for this test:
1. the query finishes sooner than 350 seconds (when it does it is around 340 seconds)
2. the query times out at 10 minutes (this is the test timeout)

I have created 2 variables for the jdbc repository (see screenshot) and pass those variables to the test script for v1 and v2 respectively. Changed the TimeoutTest to check these environment variables first and if not found to use the default ones from the test (We need to make sure we can still run the test from IntelliJ). I have used 1 at the end of the value to make sure we see it being read from the variable (see second screenshot for an example from a v1 run) 

I have also increased the timeout to 15 minutes. I don't know why some runs are taking longer than 10 seconds.

### DIFF
--- a/.github/workflows/integration-test-v1.yml
+++ b/.github/workflows/integration-test-v1.yml
@@ -104,4 +104,4 @@ jobs:
         run: chmod +x gradlew
 
       - name: Run integration tests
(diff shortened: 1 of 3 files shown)


goprean:
Screenshot 2025-04-14 at 16 55 53

goprean:
Screenshot 2025-04-14 at 16 52 22

ptiurin (approved):

PR #556, opened 2025-04-15
ci: FIR-45326 added a sandbox run for jdbc driver

goprean:
With 3.6.0 release we had broken the packdb use case. The problem was that they did not had Sfl4j library on their class path (it was not needed).

In order to simulate their use case, we could use this sandbox workflow that would do the following:
- install java (no other libraries)
- download the firebolt artifact from the run that we want to test
- create a smal test program that would connect to localhost
- make sure the workflow passes

We could add this step as part of the release workflow to make sure that we are not breaking packdb.

This is how a successful run looks like: https://github.com/firebolt-db/jdbc/actions/runs/14464007217
This is a failed run using the jar that broke packdb: https://github.com/firebolt-db/jdbc/actions/runs/14464021533

Note that the error message is the same that packdb is seeing. 

INFO: Firebolt driver version used: 3.6.0
Error: Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at com.firebolt.jdbc.service.FireboltEngineVersion2Service.<clinit>(FireboltEngineVersion2Service.java:19)
	at com.firebolt.jdbc.connection.FireboltConnectionServiceSecret.<init>(FireboltConnectionServiceSecret.java:73)
	at com.firebolt.jdbc.connection.LocalhostFireboltConnection.<init>(LocalhostFireboltConnection.java:27)
	at com.firebolt.jdbc.connection.FireboltConnectionProvider$FireboltConnectionProviderWrapper.createLocalhostFireboltConnectionServiceSecret(FireboltConnectionProvider.java:93)
	at com.firebolt.jdbc.connection.FireboltConnectionProvider.create(FireboltConnectionProvider.java:42)
	at com.firebolt.FireboltDriver.connect(FireboltDriver.java:43)
	at SmokeTest.main(SmokeTest.java:13)
Caused by: java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
	... 7 more


.github/workflows/sandbox-smoke-test.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Create java file
+        run: |
+          cat <<EOF > SmokeTest.java          
+          import com.firebolt.FireboltDriver;
+          import com.firebolt.jdbc.connection.LocalhostFireboltConnection;
+          import java.sql.Connection;
+          import java.sql.SQLException;
+          import java.util.Properties;
+          
+          public class SmokeTest {
```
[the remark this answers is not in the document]
goprean: I have no preference, as this is a small file. I have added it under src/integrationTests/resources folder.

.github/workflows/sandbox-smoke-test.yml
```diff
@@ -0,0 +1,100 @@
...
+            }
+          
+          }
+          EOF
+
+      - name: Compile class
+        run: javac -cp artifact/firebolt-jdbc-3.6.0.jar SmokeTest.java
+
+      - name: Run class
+        run: java -cp artifact/firebolt-jdbc-3.6.0.jar:. SmokeTest
```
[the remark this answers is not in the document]
goprean: Fixed.

.github/workflows/sandbox-smoke-test.yml
```diff
@@ -0,0 +1,100 @@
...
+          echo "🎯 Artifact ID: $ARTIFACT_ID"
+
+          # Download the artifact
+          mkdir -p artifact
+          gh api --method GET \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: token $GH_TOKEN" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/firebolt-db/jdbc/actions/artifacts/$ARTIFACT_ID/zip \
+            > artifact/artifact.zip
```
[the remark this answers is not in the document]
goprean: Used the download action.

ptiurin:
Looks good. Should we add it to be triggered from release workflow or integration test? I'd rather not have another manual test we have to remember to trigger.

goprean:
> Looks good. Should we add it to be triggered from release workflow or integration test? I'd rather not have another manual test we have to remember to trigger.

The way I wanted to have it is hook it up to the release.yml action in the following way:
1. trigger release.yml
   1.1 trigger build.yml (so we can upload the latest artifact and get the run id)
   1.2 get this run id (somehow) and pass it to the sandbox-smoke-test.yml (so we don't need to know what runId to pass)
   1.3. trigger integration tests 
That way one should only trigger one job for release. In the future we can have additional checks in this release flow, we can iterate on it. 

I was planning to push this in smaller steps rather than a big change review.

goprean:
> > 1.2 get this run id (somehow) and pass it to the sandbox-smoke-test.yml (so we don't need to know what runId to pass)
> 
> I don't think we need to have a run id. If the same workflow runs both upload and download then we should be able to access it, similar to how we already do it in [release](https://github.com/firebolt-db/jdbc/blob/ed322861f6e2534f3fbe1eaff07beb9fb5d774ec/.github/workflows/release.yml#L63-L66).
> 
> > I was planning to push this in smaller steps rather than a big change review.
> 
> Let's have a self-contained review here, I don't think it will be too big to add all the changes needed.

I have updated the review. The release action now does the following:
- calls the build job - make sure code compiles and tests are passing
- calls the integration job - make sure the integration tests pass
- calls the smoke test job - uses the jar that was created in the build job to run the smoke test. 

So this would be a one click process that does all the verifications that we have to do. We can add more verifications and checks later if we have to. 

Tested it here by calling the release action but not actually publishing and releasing the jar (those jobs were commented out): https://github.com/firebolt-db/jdbc/actions/runs/14516665996

.github/workflows/build.yml
```diff
@@ -14,6 +14,10 @@ on:
       sources-jar:
         description: Jar that contains only the source code
         value: ${{ jobs.build.outputs.sources-jar }}
+      run_id:
+        description: the run id of the workflow
+        value: ${{ jobs.build.outputs.run_id }}
+
```
ptiurin, line 20: Let's remove this since it's not being used in this workflow.

.github/workflows/pre-release-test.yml
```diff
@@ -0,0 +1,33 @@
...
+      FIREBOLT_STG_PASSWORD: ${{ secrets.FIREBOLT_STG_PASSWORD }}
+      FIREBOLT_CLIENT_ID_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}
+      FIREBOLT_CLIENT_SECRET_STG_NEW_IDN: ${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}
+
+  run-smoke-test:
+    needs: build
+    uses: ./.github/workflows/sandbox-smoke-test.yml
+    with:
+      jdbcJar: ${{ needs.build.outputs.uber-jar }}
+      runId: ${{ needs.build.outputs.run_id }}
```
ptiurin, line 33: `${{ github.run_id }}` should suffice here. This way we don't rely on redundant output of `build` while also keeping sandbox workflow able to run against an arbitrary runid.

goprean: Done

.github/workflows/build.yml
goprean: Done

.github/workflows/build.yml
ptiurin, line 1: Lets revert this file since nothing was really changed here. I'd prefer not adding unnecessary commits to the file history.

ptiurin (approved):
Apart from that lgtm

.github/workflows/build.yml
goprean: Done

PR #558, opened 2025-04-16
(ci) stagger nightlty runs

goprean:
I have seen this happening at least 3 times while playing with the nightly runs:

This is the latest run I have seen the error: https://github.com/firebolt-db/jdbc/actions/runs/14485636659/job/40630566975
RuntimeError: invalid command: rpc error: code = InvalidArgument desc = Name of engine integration_testing__1744782472 is not unique

We run 2 jobs in parallel. Since we have 2 variables: os and java version, I have seen that the way the jobs are scheduled first it keeps the os fix and iterate through the java versions so we will never run the same java version in parallel. 

Are adding a delay in the integration test. We are adding a delay of (java version - 5) seconds.


goprean:
> In other repos we get around this problem by [adding a suffix](https://github.com/firebolt-db/firebolt-python-sdk/blob/main/.github/workflows/nightly-v2.yml#L45-L53) to the generated db/engine name - os and language version. This way we make sure we avoid clashing with other runs and also can easily correlate a run to a database. Let's do the same here.

Done. Added the db_suffix

goprean:
Screenshot 2025-04-16 at 15 28 42
Screenshot 2025-04-16 at 15 28 06

bogdantruta-firebolt (approved):

ptiurin (approved):

PR #562, opened 2025-04-28
fix: FIR-38161 Fix failing network policy test

bogdantruta-firebolt:


src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -319,9 +319,13 @@ void unsuccessfulConnect(boolean useDatabase, boolean useEngine) {
     }
 
     private String getJdbcUrl(ConnectionInfo params, boolean useDatabase, boolean useEngine) {
+        return getJdbcUrl(params, useDatabase, useEngine, new HashMap<>());
```
goprean, line 322: if you pass the connection_cache, false as a map to this method. This is the one that actually creates the connection info. To me this is a better place to put it rather than change it here

goprean (approved):

PR #563, opened 2025-04-28
(feat) FIR-42058 add the support for statement_timeout

goprean:
When setting a timeout for a query we JDBC driver was sending the query param to the server as: max_execution_time. 
This was changed to statement_timeout so when a timeout is set on the statement the driver will add a query parameter as statement_timeout.

### DIFF
--- a/src/integrationTest/java/integration/tests/StatementTest.java
+++ b/src/integrationTest/java/integration/tests/StatementTest.java
@@ -2,22 +2,10 @@ package integration.tests;
 
 import com.firebolt.jdbc.connection.FireboltConnection;
 import com.firebolt.jdbc.exception.FireboltException;
+import com.firebolt.jdbc.testutils.TestTag;
 import integration.ConnectionInfo;
 import integration.EnvironmentCondition;
 import integration.IntegrationTest;
-import kotlin.collections.ArrayDeque;
-import lombok.CustomLog;
-import org.apache.commons.lang3.RandomStringUtils;
-import org.hamcrest.Matchers;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Tag;
-import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


bogdantruta-firebolt (approved):

src/integrationTest/java/integration/tests/StatementTest.java
```diff
@@ -558,6 +557,56 @@ void unlimitedMaxFieldSize(int maxFieldSize) throws SQLException {
...
+		try (Connection connection = createConnection(); Statement statement = connection.createStatement()) {
+			statement.setQueryTimeout(1);
+			ResultSet resultSet = statement.executeQuery("SELECT * FROM GENERATE_SERIES(1, 10000000);");
+			boolean errorFound = false;
+			while (resultSet.next()) {
+				String object = resultSet.getString(1);
+				if (object.contains("Query was canceled with reason 'Query timeout expired (1 ms)")) {
+					errorFound = true;
+				}
+			}
```
ptiurin, line 591: I think we're meant to raise a Timeout exception here

goprean: This one is a bit tricky. The spec mentions that a timeout exception should be thrown. 
But I think that would be the case if the server did not start sending any response (for example, compute a complex join query). In this case, the server starts sending some response. 
That's why I was asking in the Jira what is the expected behavior since the driver did not throw any exception even before with max_execution_time.

ptiurin: I think `max_execution_time` was never actually working on 2.0, it was ported from 1.0. Actually, just checked and it doesn't do anything on 1.0 right now either, so we shouldn't rely on the past behavior here.
Server side is separate and, as mentioned in the ticket, doesn't seem to be working correctly. However, we're concerned with the client-side behavior here and the expectation is that we throw an exception. If we can't do that with the server-side parameter then we would have to implement timeout on the client side and abort the connection when timeout is reached.

PR #567, opened 2025-05-07
Each connection should return the endpoint (should be an URL) that is connecting to

goprean:
Currently the StatementClientImpl class creates the base uri to where to post the query calls, and it constructs it by using host, port and checking the ssl on the properties. 

We already have a getEndpoint() method on the connection class that should return what is the baseURL where the firebolt is deployed. We should use this one especially since for Firebolt core we will get the endpoint from a url parameter. 

This way each connection will have the url of its deployment that can be accessed using getEndpoint().


goprean:
I am discarding this one as it seems more inter winded that I expected.

PR #569, opened 2025-05-08
(ci) FIR-44420 part2 added the integration test workflow for core

goprean:
Added the workflow to run integration tests for Firebolt core. 
The workflow takes 3 parameters:
- docker image version (we can run arm or x86 docker images)
- operating system
- java version

We are starting with ubuntu, java 11 and the latest version of x86 of the firebolt core (that was released May 8th).

Once we have a stable firebolt version we could add nightly's job that would run against that version.

### DIFF
--- a/.github/workflows/core-integration-test.yml
+++ b/.github/workflows/core-integration-test.yml
@@ -25,17 +25,16 @@ on:
         required: false
         type: string
         default: '4.21.0-pre.0.20250508101124.cfa76a685ee8_x86'
-      os_name:
-        description: 'Operating system'
-        required: false
-        type: string
-        default: 'ubuntu-latest'
       java_version:
         description: 'Java'
         required: false
         type: string
         default: '11'
-
+      os_name:
+        description: 'Operating system'
+        required: false
+        type: string
+        default: 'ubuntu-latest'
 jobs:
   run-core-integration-tests:
     runs-on: ${{ inputs.os_name }}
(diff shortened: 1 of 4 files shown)


.github/workflows/core-integration-test.yml
```diff
@@ -0,0 +1,119 @@
...
+        required: false
+        type: string
+        default: '4.21.0-pre.0.20250508101124.cfa76a685ee8_x86'
+      os_name:
+        description: 'Operating system'
+        required: false
+        type: string
+        default: 'ubuntu-latest'
+      java_version:
+        description: 'Java'
```
bogdantruta-firebolt, line 34: nit: I see that every workflow has os, java_version order, but the dispatch order is reversed, do you think we should keep the same order and description?

bogdantruta-firebolt (approved):
lgtm other than comm

.github/workflows/core-integration-test.yml
```diff
@@ -0,0 +1,119 @@
+name: Core integration tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_version:
+        description: 'The docker image tag for the firebolt core'
+        required: false
+        type: string
+        default: '4.21.0-pre.0.20250508101124.cfa76a685ee8_x86'
```
ptiurin, line 10: Is this a public image?

ptiurin (approved):
lgtm

.github/workflows/core-integration-test.yml
goprean: it is internal image. I don't think we have a released public version yet.

PR #570, opened 2025-05-08
FIR-44420 added the support firebolt core

goprean:
Added the support for Firebolt Core. 
I am planning to add tests gradually, and support. So far I have added the tests for the DatabaseMetadata. Those are basically the same tests as the ones for Cloud.


src/main/java/com/firebolt/jdbc/FireboltBackendType.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * firebolt hosted cloud version 2.0
+     */
+    CLOUD_2_0("cloud"),
+
+    /**
+     * only used for testing by packdb
+     */
+    PACKDB_DEV("packdb_dev"),
```
ptiurin, line 24: I'd call it "dev" instead. Packdb is the legacy name used internally. To avoid confusion and questions "is Firebolt build on top of some other DB" I'd keep this name out of public repos.

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
```diff
@@ -0,0 +1,141 @@
...
+                    .host(uri.getHost())
+                    .port(uri.getPort())
+                    .ssl(uri.getScheme().startsWith("https://"))
+                    .build();
+
+            // the constructor sets the httpConnectionUrl, but it checks the host and port. For Core the httpConnectionUrl should be the url parameter.
+            this.httpConnectionUrl = sessionProperties.getUrl();
+        } catch (URISyntaxException e) {
+            // this should not happen as we had validate the url already validated
+        }
```
ptiurin, line 52: We should still raise here. Right now we validate it but if later on we refactor and forget the validation it's handy to catch this early.

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
```diff
@@ -0,0 +1,141 @@
...
+            int port = fireboltCoreUrl.getPort();
+
+            if (StringUtils.isEmpty(protocol) || StringUtils.isEmpty(host)) {
+                throw new SQLException("Invalid URL format. URL must be in the form: <protocol>://<host>:<port>");
+            }
+
+            // Validate port range
+            if (port <= 0 || port > 65535) {
+                throw new SQLException("Invalid URL format. URL must be in the form: <protocol>://<host>:<port>");
+            }
```
ptiurin, line 84: Can we be more specific here in the error? "Port value is out of valid range" or something.

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
```diff
@@ -0,0 +1,141 @@
...
+                if (isIPv6) {
+                    hostToValidate = host.substring(1, host.length() - 1);
+                }
+
+                InetAddressValidator inetAddressValidator = InetAddressValidator.getInstance();
+                DomainValidator domainValidator = DomainValidator.getInstance(false);
+
+                if (!inetAddressValidator.isValid(hostToValidate) && !domainValidator.isValid(hostToValidate)) {
+                    throw new SQLException("Invalid URL format. URL must be in the form: <protocol>://<host>:<port>");
+                }
```
ptiurin, line 100: Same here, we can probably specify that host is invalid specifically. Maybe the user specified it in `<protocol>://<host>:<port>` format so this error will be confusing.

ptiurin (commented):
Apart from that looks good!

src/main/java/com/firebolt/jdbc/FireboltBackendType.java
goprean: Done. Changed it to dev.

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
goprean: Done

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
goprean: Done

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
goprean: Done.

ptiurin (approved):

PR #573, opened 2025-05-13
(ci) FIR-44420-added-nightlys-for-core

goprean:
When running nightly's we will try to fetch the latest docker image available for x86 and version 4.21 (this is the current version of packdb that is being developed). 
We will then call the core-integration-tests to run the core integration tests. 

This way we will catch if there are some bugs introduced by the firebolt core.

### DIFF
--- a/.github/workflows/nightly-core.yml
+++ b/.github/workflows/nightly-core.yml
@@ -1,0 +1,77 @@
+name: Nightly integration tests core
+
+on:
+  schedule:
+    - cron: '30 0 * * *'        # Daily at 03:00 UTC
+  workflow_dispatch:             # Allow manual trigger
+
+jobs:
+  detect-latest-image-available:
+    runs-on: ubuntu-latest
+    outputs:
+      latest_tag: ${{ steps.set-tag.outputs.latest_tag }}
+    steps:
+      - name: detect latest image availabe
+        id: set-tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PREFIX: "4.21"
+          SUFFIX: "x86"
+        run: |
+          echo "Querying GHCR image tags from $OWNER/$IMAGE_NAME..."
+          
+          # Query image versions from GitHub Container Registry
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-core.yml
```diff
@@ -0,0 +1,77 @@
...
+        id: set-tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PREFIX: "4.21"
+          SUFFIX: "x86"
+        run: |
+          echo "Querying GHCR image tags from $OWNER/$IMAGE_NAME..."
+          
+          # Query image versions from GitHub Container Registry
+          response=$(curl -s -H "Authorization: Bearer $GH_TOKEN" \
```
ptiurin, line 24: You can use `${{ secrets.GITHUB_TOKEN }}` directly here, without passing an env variable.

.github/workflows/nightly-core.yml
```diff
@@ -0,0 +1,77 @@
...
+      latest_tag: ${{ steps.set-tag.outputs.latest_tag }}
+    steps:
+      - name: detect latest image availabe
+        id: set-tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PREFIX: "4.21"
+          SUFFIX: "x86"
+        run: |
+          echo "Querying GHCR image tags from $OWNER/$IMAGE_NAME..."
```
ptiurin, line 21: I don't think we have `$OWNER/$IMAGE_NAME` defined

ptiurin (approved):
Apart from that lgtm. I suppose we'll eventually move to `:latest` so this will become obsolete

.github/workflows/nightly-core.yml
goprean: Done.

.github/workflows/nightly-core.yml
goprean: Ups.. fixed it. Thanks for catching it.

PR #575, opened 2025-05-16
fix: FIR-44413 ResultSet isBefore/isLast/isAfterLast fix

bogdantruta-firebolt:


src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -288,6 +288,43 @@ void shouldNotBeLastWhenThereIsMoreData() throws SQLException {
 
 	@Test
 	void shouldNotBeLastAtLastLine() throws SQLException {
+		inputStream = getInputStreamWithCommonResponseExample();
```
ptiurin, line 291: Name of the test here is confusing. Perhaps `shouldNotBeLastIfNotLast`?

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -288,6 +288,43 @@ void shouldNotBeLastWhenThereIsMoreData() throws SQLException {
...
+		}
+		assertEquals(2, count);
+		assertFalse(resultSet.isBeforeFirst());
+		assertFalse(resultSet.isLast());
+		assertTrue(resultSet.isAfterLast());
+	}
+
+	@Test
+	void shouldBeAtLastLine() throws SQLException {
 		inputStream = getInputStreamWithCommonResponseExample();
```
ptiurin, line 328: `shouldBeLastAtLastLine`?

ptiurin (approved):
otherwise lgtm

src/main/java/com/firebolt/jdbc/resultset/FireboltResultSet.java
```diff
@@ -447,7 +450,7 @@ public Object getObject(String column) throws SQLException {
 	@Override
 	public boolean isBeforeFirst() throws SQLException {
 		checkStreamNotClosed();
-		return currentRow < 3 || !hasNext();
+		return currentRow < 3;
```
goprean, line 453: This is the java doc  for this isBeforeFirstMethod()

'Returns:
true if the cursor is before the first row; false if the cursor is at any other position or the result set contains no rows'

So to me this should be true only if: currentRow == 2 && nextLine != null (so if we had read the first 2 header lines) and if there is an actual result. Everything else should return false.

goprean (approved):

src/test/java/com/firebolt/jdbc/resultset/FireboltResultSetTest.java
```diff
@@ -287,7 +287,62 @@ void shouldNotBeLastWhenThereIsMoreData() throws SQLException {
...
+		assertFalse(resultSet.isLast());
+		assertFalse(resultSet.next());
+		assertFalse(resultSet.isLast());
+	}
+
+	@Test
+	void shouldNotBeforeFirst() throws SQLException {
+		inputStream = getInputStreamWithNoRows();
+		resultSet = createResultSet(inputStream);
+		assertFalse(resultSet.isBeforeFirst());
```
goprean, line 312: let's also check isLast here as well, I see that was changed as well.

src/test/resources/responses/firebolt-response-with-no-results
```diff
@@ -0,0 +1,2 @@
+1
```
goprean, line 1: even when we do Select 1 this is the format of the result? I mean even in that case we do return 2 header lines? Just double check.

goprean (approved):

src/test/resources/responses/firebolt-response-with-no-results
bogdantruta-firebolt: if we do select one we also get a result, the 1. I tried with a select from a table with a where clause which returned 0 rows and yes, we do get the headers.

goprean: So anytime we get a result back we get the header lines. That's all good then.

PR #577, opened 2025-05-21
(test) FIR-44420 - include tests tags rather than exclude them when running integration tests

goprean:
Before firebolt core we had 2 backend: v1 and v2 and we were controlling the tagging of the tests using the following tags:
- common (would run against both backends)
- v1 - would run only on v1
- v2 - would run only on v2

We used to run the integration tests for v1 by excluding the test with v2: -DexcludeTags=v2

This approach does not work when we add the core backend as well since now we have 7 possibilities instead of 3: 
- test should run against all backends
- test should run against just one backend (v1, v2 or core)
- tests should run on combination of backends (v1 and v2 but not on core, v1 and core but not v2, v2 and core but not v1)

Changed the tests to always specify the backends that the test should run against. Introduced another abstract class CommonIntegrationTest that can be extended when the tests in that class should run against all backends. 

Changed the command line for each integration tests to include the backend that it needs to run agasint: -DincludeTags=core (would run the tests annotated with core)

NOTE: there are 5 tests that when run against core are failing. I will address the fixes for them in the next PR

### DIFF
--- a/src/integrationTest/java/integration/GlobalSetupExtension.java
+++ b/src/integrationTest/java/integration/GlobalSetupExtension.java
@@ -1,4 +1,3 @@
-
 package integration;
 
 import com.firebolt.jdbc.FireboltBackendType;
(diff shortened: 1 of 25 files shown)


ptiurin (approved):
lgtm

src/integrationTest/java/integration/GlobalSetupExtension.java
```diff
@@ -57,43 +56,43 @@ public void beforeAll(ExtensionContext context) throws SQLException {
...
 
     /**
-     * The way that our integration workflow are setup is to use the -DexcludeTag to exclude the version of backend that we don't want in the current run.
+     * The way that our integration workflow are set up is to use the -DincludeTags to specify the tests groups to run (of which we should include one of: v1, v2 or core) to know against which backend the tests are running
      * @return
      */
-    private Set<String> getExcludeTags() {
-        String excludeTags = System.getProperty("excludeTags", "");
+    private Set<String> getIncludeTags() {
+        String excludeTags = System.getProperty("includeTags", "core");
```
bogdantruta-firebolt, line 95: the variable should be includeTags

src/integrationTest/java/integration/tests/ConnectionTest.java
```diff
@@ -159,7 +159,7 @@ void successfulConnect(boolean useDatabase, boolean useEngine) throws SQLExcepti
     }
 
     @Test
-    @Tag("v2")
+    @Tag(TestTag.V2)
     @Tag("slow")
```
bogdantruta-firebolt, line 163: could you also change the "slow" value to the testTag?

src/integrationTest/java/integration/tests/PreparedStatementTest.java
```diff
@@ -136,6 +140,9 @@ Stream<Arguments> numericTypes() {
 
 	@ParameterizedTest(name = "{0}")
 	@MethodSource("numericTypes")
+	@Tag(TestTag.V1)
+	@Tag(TestTag.V2)
+	@Tag(TestTag.CORE)
```
bogdantruta-firebolt, line 145: Do you think it could be a good idea to extract test that should run on all backends into a separate class that would extend CommonIntegrationTest?

src/integrationTest/java/integration/tests/SpecialValuesTest.java
```diff
@@ -168,13 +183,16 @@ void nanDoubleSystemEngine(String query) throws SQLException {
     @ValueSource(strings = {
             "select 'nan'::float", "select '+nan'::float", "select '-nan'::float",
     })
-    @Tag("v2")
+    @Tag(TestTag.V2)
```
bogdantruta-firebolt, line 186: How about this as well? Maybe here you could extract the specific V2 special values into a separate test class

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
```diff
@@ -44,21 +45,23 @@ void disconnect() throws SQLException {
...
-    @Tag("v2")
+    @Tag(TestTag.V2)
     void getSchemas() throws SQLException {
         String database = integration.ConnectionInfo.getInstance().getDatabase();
         assertEquals(List.of(List.of("information_schema", database)), getRows(DatabaseMetaData::getSchemas));
     }
 
 	@Test
-	@Tag("v2")
+    @Tag(TestTag.V2)
```
bogdantruta-firebolt, line 64: Indentation seems weird here

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
```diff
@@ -44,21 +45,23 @@ void disconnect() throws SQLException {
...
     @Test
     void readOnly() throws SQLException {
         assertFalse(connection.isReadOnly());
         assertFalse(connection.getMetaData().isReadOnly());
     }
 
     @Test
-    @Tag("v2")
+    @Tag(TestTag.V2)
```
bogdantruta-firebolt, line 57: nit: Could we settle on an order of annotations? Test after Tags, or the other way around?

src/integrationTest/java/integration/GlobalSetupExtension.java
goprean: Fixed.

src/integrationTest/java/integration/tests/ConnectionTest.java
goprean: Done.

src/integrationTest/java/integration/tests/PreparedStatementTest.java
goprean: In my opinion it would be easier to have it in one file the test that are related to one feature (like prepared statement) since it might be harder to know which test is where (in the common or the specific one). My preference is to keep one file.

src/integrationTest/java/integration/tests/SpecialValuesTest.java
goprean: Same as above. Plus on the ones that we with v2 only I need to double check that they cannot run on core. I think the ones with the user connection might be able to run on core as well. The ones with system engine, definitely not.

src/integrationTest/java/integration/tests/SystemEngineDatabaseMetaDataTest.java
ptiurin: If we decide on any code style items, let's make sure we're automatically apply them instead of having to think about each one before/during the review.
In Python and some other repos we have a linter that runs on pre commit. Something similar would be great here, or in the meantime a shared IDE settings file will suffice.

stepansergeevitch (approved):

goprean:
Screenshot 2025-05-21 at 14 24 33

src/integrationTest/java/integration/tests/PreparedStatementTest.java
bogdantruta-firebolt: Would it be an option to move the tags that are present on all test to the class level and keep uncommon ones for each test? (ex v1 for each test and v2/core for the whole class)

PR #579, opened 2025-05-23
(feat) FIR-44420 part 6 - Run tests against the https

goprean:


src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
```diff
@@ -78,13 +85,71 @@ void testMissingUrl() {
         assertTrue(exception.getMessage().contains("Url is required for firebolt core"));
     }
 
+    @Test
+    void canConnectToCoreWithoutSpecifyingADB() throws SQLException {
```
bogdantruta-firebolt, line 89: nit: can you make the b lowercase?

src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
```diff
@@ -78,13 +85,71 @@ void testMissingUrl() {
...
+            assertEquals("localhost", fireboltProperties.getHost());
+            assertEquals(3473, fireboltProperties.getPort());
+            assertEquals("my_db", fireboltProperties.getDatabase());
+        }
+    }
+
+    @Test
+    void canConnectOverHttps() throws SQLException {
+        Map<String, String> connectionParams = Map.of(
+                "url", "https://localhost:3473",
```
bogdantruta-firebolt, line 121: the port should be 443 for https?

src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
```diff
@@ -78,13 +85,71 @@ void testMissingUrl() {
...
+            assertEquals(3473, fireboltProperties.getPort());
+            assertTrue(StringUtils.isBlank(fireboltProperties.getDatabase()));
+        }
+
+    }
+
+    @Test
+    void canConnectOverHttp() throws SQLException {
+        Map<String, String> connectionParams = Map.of(
+                "url", "http://localhost:3473",
```
bogdantruta-firebolt, line 106: could we extract the http and https urls to constants? maybe also the ports?

src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
```diff
@@ -78,13 +85,71 @@ void testMissingUrl() {
...
+
+        String params = parameters.entrySet().stream()
+                        .map(entry -> entry.getKey() + "=" + entry.getValue())
+                        .collect(Collectors.joining("&"));
+
+        jdbcUrlBuilder.append("&").append(params);
+
+        return new FireboltCoreConnection(jdbcUrlBuilder.toString(), new Properties());
+    }
+}
```
bogdantruta-firebolt, line 155: nit: new line at eof

src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
goprean: it does not really matter as this is how I have set it up. but we could set it up for any port since this is just the unit test.

src/test/java/com/firebolt/jdbc/connection/FireboltCoreConnectionTest.java
goprean: I think it would be easier to read like this when you see the whole URL. In tests I don't mind duplication for readability, unless Sonar complains.

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
```diff
@@ -42,7 +42,7 @@ protected void authenticate() throws SQLException {
             sessionProperties = sessionProperties.toBuilder()
                     .host(uri.getHost())
                     .port(uri.getPort())
-                    .ssl(uri.getScheme().startsWith("https://"))
+                    .ssl(uri.getScheme().startsWith("https"))
```
ptiurin, line 45: Why is this changing, do we not support slashes?

ptiurin (approved):
One question, otherwise lgtm

src/main/java/com/firebolt/jdbc/connection/FireboltCoreConnection.java
goprean: we do support // but when the uri is parsed, the uri.getScheme() will return just "https" . So it was always defaulting to false.

stepansergeevitch (approved):

PR #580, opened 2025-05-23
(test) FIR-44420 fixed the failing tests against 4.22

goprean:
Core team has release the fixes in 4.22. Before it was 4.21.

They have fixed some issues with the setting of parameters not being recognized. 

There are few other changes as well:
- force_pgdate_timestampntz is not supported any more for core (actually it is not doing anything for cloud as well)
- time_zone setting is not supported for core but the new property is timezone. This is the same for cloud but it seems that on cloud both work. 
- error messages for an unknown parameter is different if the query is ran against user engine or system engine. So changed the tests to reflect that. For core it will be similar to cloud when running on user engine.

### DIFF
--- a/src/integrationTest/java/integration/tests/TimestampTest.java
+++ b/src/integrationTest/java/integration/tests/TimestampTest.java
@@ -1,7 +1,8 @@
 package integration.tests;
 
 import com.firebolt.jdbc.testutils.AssertionUtil;
-import integration.CommonIntegrationTest;
+import com.firebolt.jdbc.testutils.TestTag;
+import integration.IntegrationTest;
 import io.zonky.test.db.postgres.embedded.EmbeddedPostgres;
 import java.io.IOException;
 import java.sql.Connection;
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/integration/tests/TimestampTest.java
```diff
@@ -210,8 +294,29 @@ void shouldReturnTimestampFromTimestampTzWithTzWithHoursAndMinutes() throws SQLE
...
+			assertEquals(expectedTimestamp, resultSet.getObject(1));
+			compareAllDateTimeResultSetValuesWithPostgres(resultSet, "SELECT '1111-01-05 17:04:42.123456'::timestamptz",
+					"Asia/Calcutta");
+		}
+	}
+
+	@Tag(TestTag.V1)
+	@Tag(TestTag.V2)
+	@Test
+	void shouldReturnTimestampFromTimestampTzWithTzWithHoursAndMinutesAndSecondsWhenUsingTime_ZoneSetting() throws SQLException {
```
stepansergeevitch, line 319: Can we also test `timezone` with no `_` + CORE?

PR #582, opened 2025-05-26
(test) FIR-44420 Added more tests for core

goprean:
Added the CORE annotation to 4 additional tests. 

Also added one more test that verifies that certain statements are not valid for core: USE ENGINE <abc> is not valid for core.


bogdantruta-firebolt (approved):
lgtm

PR #583, opened 2025-05-27
Allow multiple nightly runs for different version of prefixes

goprean:
Going forward we might support multiple docker image versions (4.22-xxx, 4.23.xxx, etc)
Changed the nightly's to get the version prefixes from a GitHub variable so we can change them easily without changing the code. 

Going forward I will remove the 4.21 since I think we will release the 4.22 but you can see in the screenshot how the GitHub variable looks like.


goprean:
Screenshot 2025-05-27 at 10 30 08

This is an example of the run: https://github.com/firebolt-db/jdbc/actions/runs/15257815705

stepansergeevitch (approved):

PR #608, opened 2025-07-15
 (ci) FIR-47600 - Use the preview-rc tag to look for the core docker image to run the nightly's

goprean:
The latest image that the customers can download, until we go GA, is tagged with preview-rc.
Instead of using the latest docker image available (since there might be image published for testing or for specific customer), we will always run nightly's against the image with preview-rc. When the Core team wants to release a new docker image, it will tag is with preview-rc and the next nightly run will pick it up. 

I had to remove the -ti option when running docker as there is no TTY for GitHub actions. I was getting this error:
`the input device is not a TTY`


.github/workflows/nightly-core.yml
```diff
@@ -29,56 +30,71 @@ jobs:
...
+          VERSION_TAG=$(echo "$response" | jq -r --arg tag "$TAG" '
+            if . == null then
+              error("Response is null")
+            else
+              .[] | 
+              if .metadata == null then
+                empty
+              elif .metadata.container == null then
+                empty
+              elif .metadata.container.tags == null then
```
ptiurin, line 43: Do we need to do this stuff anymore? Can we just do an equivalent of `docker pull firebolt-core:preview-rc`?

goprean: I wanted to pull the specific image for a specific architecture. So in this case amd64. 
We have dockerImageVersion:preview-rc.
and I wanted to pull dockerImageVersion-amd64

goprean: I have removed the image tag detection and only use the preview-rc tag for the nightly.

PR #613, opened 2025-07-23
Fir 47755 migrate jdbc publishing to maven central

stepansergeevitch:
Added a manual upload step for a package from Maven docs: 
https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#ensuring-deployment-visibility-in-the-central-publisher-portal

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,3 +93,9 @@ jobs:
         MAVEN_REPO_PASSWORD: ${{ secrets.MAVEN_REPO_PASSWORD }}
         GRADLE_SIGNING_KEY: ${{ secrets.GRADLE_SIGNING_KEY }}
         GRADLE_SIGNING_PASSWORD: ${{ secrets.GRADLE_SIGNING_PASSWORD }}
+
+    - name: Trigger manual package upload via API
+      run: ./gradlew manualUploadToMavenCentral
+      env:
+        MAVEN_REPO_USERNAME: ${{ secrets.MAVEN_REPO_USERNAME }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -78,18 +78,24 @@ jobs:
...
         name: ${{ needs.build.outputs.sources-jar }}
-    - uses: xresloader/upload-to-github-release@v1
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        file: ${{ needs.build.outputs.uber-jar }}
-        tags: true
-        draft: false
-        tag_name: ${{ github.event.inputs.tag_name }}
+#    - uses: xresloader/upload-to-github-release@v1
```
goprean, line 81: this seems to still be needed. It uploads the jar to the release

gradle.properties
```diff
@@ -1,2 +1,2 @@
-version=3.6.7
+version=3.6.6
```
goprean, line 1: is this for testing? Since we are on 3.6.7.

.github/workflows/release.yml
stepansergeevitch: This was just for testing

goprean (approved):

goprean (commented):
Let's also change the release process here, since there are steps that are not needed anymore. https://packboard.atlassian.net/wiki/spaces/INTERFACES/pages/3458662503/Releasing+and+publishing#Publishing-to-OSSRH-Staging

PR #623, opened 2025-08-21
(feat) FIR-48534 expose the exception type of 413 when payload too big

goprean:
There are cases where we need to know more about the SqlException that happens. We are adding the ExceptionType value to the exception. We missed one case when the payload entity is too big. There are clients that will need to distinguish this exception (send less prepared statements or have the body smaller)


src/main/java/com/firebolt/jdbc/exception/ExceptionType.java
```diff
@@ -4,6 +4,12 @@
  * This class represents the types of exceptions that may be thrown
  */
 public enum ExceptionType {
-	ERROR, UNAUTHORIZED, TYPE_NOT_SUPPORTED, TYPE_TRANSFORMATION_ERROR, RESOURCE_NOT_FOUND, CANCELED, INVALID_REQUEST,
-	TOO_MANY_REQUESTS
+	ERROR, UNAUTHORIZED,
```
bogdantruta-firebolt, line 7: maybe also split these onto 2 different lines for ease of view?

bogdantruta-firebolt (approved):

src/main/java/com/firebolt/jdbc/exception/ExceptionType.java
goprean: Done.

bogdantruta-firebolt (approved):

PR #626, opened 2025-08-26
(ci) FIR-48736 automate the release of the artifact from maven staging repository

goprean:
In our release flow there is still one manual step: 
```
Login to Maven Central:
Go to [Maven Central](https://central.sonatype.com/) and log in with Maven credentials. They can be found in Keeper or requested if you don't have them yet
Find the deployment:
Go to [Deployments](https://central.sonatype.com/publishing) and find the uploaded package
Publish the deployment:
Click on the Publish button near the deployment. Wait for the status to change from Publishing to Published (takes ~10 minutes).
```

Even this step should be automated. We can do this by using this plugin: https://plugins.gradle.org/plugin/io.github.gradle-nexus.publish-plugin that will be called at the end of the release workflow.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -344,10 +344,10 @@ publishing {
 nexusPublishing {
     repositories {
         sonatype {
-            nexusUrl = uri('https://s01.oss.sonatype.org/service/local/')
-            snapshotRepositoryUrl = uri('https://s01.oss.sonatype.org/content/repositories/snapshots/')
-            username = System.getenv('MAVEN_REPO_USERNAME')
-            password = System.getenv('MAVEN_REPO_PASSWORD')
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -339,6 +340,18 @@ publishing {
     }
 }
 
+// Automate close and release of Sonatype staging repository
+nexusPublishing {
+    repositories {
+        sonatype {
+            nexusUrl = uri('https://s01.oss.sonatype.org/service/local/')
```
stepansergeevitch, line 347: I believe this is an ossrh link, which has been deprecated. We're now uploading to maven central portal, see https://central.sonatype.org/pages/ossrh-eol/

goprean: You are right. Changed it.

stepansergeevitch (approved):

PR #629, opened 2025-08-27
test (noTicket) added some tests for boolean prepared statement

goprean:
Added some variations of how we can insert true or false for a column using prepared statements. Also added negative test cases for the boolean values that are rejected.

### DIFF
--- a/src/integrationTest/java/integration/tests/datatype/BooleanTest.java
+++ b/src/integrationTest/java/integration/tests/datatype/BooleanTest.java
@@ -1,0 +1,215 @@
+package integration.tests.datatype;
+
+import com.firebolt.jdbc.exception.FireboltException;
+import com.firebolt.jdbc.testutils.TestTag;
+import integration.IntegrationTest;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag(TestTag.CORE)
+@Tag(TestTag.V2)
+public class BooleanTest extends IntegrationTest {
+
+    @BeforeEach
+    void beforeEach() throws SQLException {
+        // Create a test table for the large data insertion test
+        try (Connection connection = createConnection()) {
+            connection.createStatement().execute("DROP TABLE IF EXISTS boolean_table");
+            connection.createStatement().execute("CREATE TABLE IF NOT EXISTS boolean_table (id INT, value BOOLEAN)");
+        }
+    }
+
+    @AfterEach
+    void afterEach() throws SQLException {
+        // drop the table
+        try (Connection connection = createConnection()) {
+            connection.createStatement().execute("DROP TABLE IF EXISTS boolean_table");
+        }
+    }
+
+    @Test
+    void canUsePreparedStatementsToInsertTrueValueInBooleanColumn() throws SQLException {
+        try (Connection connection = createConnection();
+             PreparedStatement preparedStatement = connection.prepareStatement("INSERT into boolean_table(id, value) values(?,?);");
+             Statement verificationStatement = connection.createStatement()) {
+            // int value 1 is considered true
+            preparedStatement.setInt(1, 1);
+            preparedStatement.setString(2, "1");
+            preparedStatement.addBatch();
+
+            // any positive integer is also considered true
+            preparedStatement.setInt(1, 2);
+            preparedStatement.setString(2, "456");
+            preparedStatement.addBatch();
+
+            // even using set object with any positive integer is considered true
+            preparedStatement.setInt(1, 3);
+            preparedStatement.setObject(2, "789");
+            preparedStatement.addBatch();
+
+            // true, t or any case of true is considered true (TRUE, truE, True, etc)
+            preparedStatement.setInt(1, 4);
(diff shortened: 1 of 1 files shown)


src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Test
+    void canUsePreparedStatementsToInsertTrueValueInBooleanColumn() throws SQLException {
+        try (Connection connection = createConnection();
+             PreparedStatement preparedStatement = connection.prepareStatement("INSERT into boolean_table(id, value) values(?,?);");
+             Statement verificationStatement = connection.createStatement()) {
+            // int value 1 is considered true
+            preparedStatement.setInt(1, 1);
+            preparedStatement.setString(2, "1");
```
bogdantruta-firebolt, line 48: I think this adds the value '1' in the query which is a string, not an int

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+            preparedStatement.addBatch();
+
+            // any positive integer is also considered true
+            preparedStatement.setInt(1, 2);
+            preparedStatement.setString(2, "456");
```
bogdantruta-firebolt, line 53: same here. setInt would work better I think

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+            preparedStatement.addBatch();
+
+            // upper case T
+            preparedStatement.setInt(1, 8);
+            preparedStatement.setString(2, "T");
+            preparedStatement.addBatch();
+
+            // even a double would still be considered as true
+            preparedStatement.setInt(1, 9);
+            preparedStatement.setString(2, "0.1123");
```
bogdantruta-firebolt, line 87: this is also a string

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Test
+    void canUsePreparedStatementsToInsertFalseValueInBooleanColumn() throws SQLException {
+        try (Connection connection = createConnection();
+             PreparedStatement preparedStatement = connection.prepareStatement("INSERT into boolean_table(id, value) values(?,?);");
+             Statement verificationStatement = connection.createStatement()) {
+
+            preparedStatement.setInt(1, 1);
+            preparedStatement.setString(2, "0");
```
bogdantruta-firebolt, line 115: same here and all other setStrings for setting ints/floats/doubles

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+            preparedStatement.setBoolean(2, true);
+            preparedStatement.addBatch();
+
+            preparedStatement.executeBatch();
+
+            ResultSet resultSet = verificationStatement.executeQuery("select * from boolean_table");
+            int actualCount = 0;
+            while (resultSet.next()) {
+                actualCount++;
+                assertEquals(true, resultSet.getBoolean(2));
```
bogdantruta-firebolt, line 100: could use assertTrue

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+            preparedStatement.setBoolean(2, false);
+            preparedStatement.addBatch();
+
+            preparedStatement.executeBatch();
+
+            ResultSet resultSet = verificationStatement.executeQuery("select * from boolean_table");
+            int actualCount = 0;
+            while (resultSet.next()) {
+                actualCount++;
+                assertEquals(false, resultSet.getBoolean(2));
```
bogdantruta-firebolt, line 162: could use assertFalse

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
```diff
@@ -0,0 +1,215 @@
...
+    @Test
+    void willFailToSetTheBooleanValue() throws SQLException {
+        try (Connection connection = createConnection();
+             PreparedStatement preparedStatement = connection.prepareStatement("INSERT into boolean_table(id, value) values(?,?);");
+             Statement verificationStatement = connection.createStatement()) {
+
+            // int value is of 0 or 1 cannot be converted to boolean
+            preparedStatement.setInt(1, 1);
+            preparedStatement.setInt(2, 0);
+            assertThrows(FireboltException.class, () -> preparedStatement.execute());
```
bogdantruta-firebolt, line 179: can use preparedStatement::execute everywhere here

bogdantruta-firebolt (commented):
Based on docs, all types need explicit casting to boolean, but apparently text can be cast directly. Int values like one from the throws test would work with explicit casting.
https://docs.firebolt.io/reference-sql/data-types#data-types

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
goprean: setInt would not work. See the negative tests cases. It is a bit weird.

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
goprean: Done

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
goprean: done

src/integrationTest/java/integration/tests/datatype/BooleanTest.java
bogdantruta-firebolt: I eventually saw that but indeed, it's a weird behavior

bogdantruta-firebolt (approved):

PR #631, opened 2025-08-29
(test) FIR-48811 - added a test that fails for bigint

goprean:
When the bug FIR-48811 is resolved this test should be enabled.


bogdantruta-firebolt (approved):
maybe add a reference to the task number in the test class, near the disabled annotation, other than that lgtm

bogdantruta-firebolt (approved):

PR #638, opened 2025-09-30
(test): fix the test that was failing due to the engine url caching added on the backend

goprean:
This test that was fixed can fail with 2 different error messages depending on where it fails. If the engine URL is in cache it fails sooner with account name error, if not it fails when trying to use the database with the error message that the database is not authorized.

### DIFF
--- a/src/integrationTest/java/integration/tests/SystemEngineTest.java
+++ b/src/integrationTest/java/integration/tests/SystemEngineTest.java
@@ -261,7 +261,13 @@ public class SystemEngineTest extends IntegrationTest {
 
 				((CacheListener)connection).cleanup();
 				SQLException e = assertThrows(SQLException.class, () -> DriverManager.getConnection(jdbcUrl, clientId, clientSecret));
-				assertTrue(e.getMessage().matches(format("Account '%s' does not exist in this organization or is not authorized.+RBAC.+", current.getAccount())), "Unexpected exception message: " + e.getMessage());
+
+				// There could be 2 error messages that are being thrown by this call. There has been caching introduced in the backend for the engine URL (FIR-49039)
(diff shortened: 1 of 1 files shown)


src/integrationTest/java/integration/tests/SystemEngineTest.java
```diff
@@ -261,7 +261,13 @@ void connectToAccountWithoutUser() throws SQLException {
...
 				SQLException e = assertThrows(SQLException.class, () -> DriverManager.getConnection(jdbcUrl, clientId, clientSecret));
-				assertTrue(e.getMessage().matches(format("Account '%s' does not exist in this organization or is not authorized.+RBAC.+", current.getAccount())), "Unexpected exception message: " + e.getMessage());
+
+				// There could be 2 error messages that are being thrown by this call. There has been caching introduced in the backend for the engine URL (FIR-49039)
+				// in case the engine url is not in the cache then the connection establishment fails with error1.
+				// in case the engine url is in cache, then the fetching of the engine url is fine but fails when trying to use the database (thus error2)
+				String errorMessage1 = format("Account '%s' does not exist in this organization or is not authorized.+RBAC.+", current.getAccount());
+				String errorMessage2 = format("Database '%s' does not exist or not authorized.", current.getDatabase());
+				assertTrue(e.getMessage().matches(errorMessage1) || e.getMessage().contains(errorMessage2), "Unexpected exception message: " + e.getMessage());
 			} finally {
```
ptiurin, line 271: We should just clear cache here instead of making the test nondeterministic. Either that or having two tests that test both cases.

ptiurin (approved):

bogdantruta-firebolt (approved):

PR #742, opened 2026-04-21
fix(metadata): return correct display size for all data types (FB-762)

moshap-firebolt:
**Background**

FB-762: getColumnDisplaySize() was hardcoded to return 80 for all column types, causing BI tools like Sisense to truncate TEXT column values.

**Summary**

Implemented proper display size calculation for all Firebolt data types, following the PostgreSQL JDBC driver approach:

Created new TypeDisplaySize utility class that computes display sizes based on data type For unlimited types (TEXT, BYTEA, ARRAY, STRUCT, JSON, GEOGRAPHY), returns Integer.MAX_VALUE For numeric types, returns PostgreSQL-compatible values (e.g., INTEGER → 11, BIGINT → 20) For date/time types, returns format-based sizes (DATE → 13, TIMESTAMP → 26, TIMESTAMPTZ → 32) For NUMERIC with precision, returns 1 + precision + (scale > 0 ? 1 : 0) Modified FireboltResultSetMetaData.getColumnDisplaySize() to delegate to the new utility Display size reference:

Type	Display Size
BOOLEAN	1
INTEGER	11
BIGINT	20
REAL	15
DOUBLE PRECISION	25
TEXT (unbounded)	Integer.MAX_VALUE
DATE	13
TIMESTAMP	26
TIMESTAMPTZ	32
NUMERIC(p,s)	1 + p + (s>0 ? 1 : 0)
BYTEA/ARRAY/STRUCT/JSON/GEOGRAPHY	Integer.MAX_VALUE

**Test Plan**

Added TypeDisplaySizeTest with comprehensive coverage for all data types Updated FireboltResultSetMetaDataTest to verify correct delegation and values All existing tests pass

### DIFF
--- a/src/main/java/com/firebolt/jdbc/resultset/FireboltResultSetMetaData.java
+++ b/src/main/java/com/firebolt/jdbc/resultset/FireboltResultSetMetaData.java
@@ -2,6 +2,8 @@ package com.firebolt.jdbc.resultset;
 
 import com.firebolt.jdbc.GenericWrapper;
 import com.firebolt.jdbc.resultset.column.Column;
+import com.firebolt.jdbc.resultset.column.ColumnType;
+import com.firebolt.jdbc.type.TypeDisplaySize;
 
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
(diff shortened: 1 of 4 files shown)


freitmi (approved):

steve751:
what about bounded text, eg varchar(20) all seem to be returning Integer.MAX_VALUE?

moshap-firebolt:
> what about bounded text, eg varchar(20) all seem to be returning Integer.MAX_VALUE?

@steve751 - Firebolt ignores the bound, so `varchar(20)` behaves the same as unbounded `text`

steve751:
@moshap-firebolt ignoring the bound isn't how other jdbcs behave this seems to only partially fix the issue. are you able to return the bound?

moshap-firebolt:
> @moshap-firebolt ignoring the bound isn't how other jdbcs behave this seems to only partially fix the issue. are you able to return the bound?

Firebolt doesn't track or enforce these bounds in the type system, so most of the work would be on the engine side. After that exposing them through JDBC should be straightforward.
(note, that systems like Postgres only preserve these bounds for simple operators, like straight projection in SELECT or GROUP BY - it is lost in most other operators and functions)

So for now I am going to merge this PR, as this is already an improvement over current state.

